PR #17, opened 2026-06-17
fix: deliver subtitle streams as External with DeliveryUrl

Bobbls:
Subtitle streams were only marked as Embed delivery with no DeliveryUrl, so Jellyfin-web could not fetch or render them. This matches Jellyfin's PlaybackInfo behaviour:

- Set DeliveryUrl on every subtitle stream pointing to the extraction endpoint (/Videos/{id}/{source}/Subtitles/{idx}/0/Stream.{format})
- Change delivery_method from Embed to External so web clients know to fetch from the URL
- Support Stream.js format (Jellyfin JSON TrackEvents) in the extraction endpoint
- Cache extracted SRT to disk so ffmpeg only runs once per stream
- Add -copyts to ffmpeg extraction to preserve subtitle timestamps

### DIFF
--- a/crates/remux-server/src/api/playback.rs
+++ b/crates/remux-server/src/api/playback.rs
@@ -414,36 +414,49 @@ async fn items_playbackinfo_inner(
         let effective_sub_idx = q
             .subtitle_stream_index
             .or(source.default_subtitle_stream_index);
-        let needs_pgs_burn = effective_sub_idx.map_or(false, |idx| {
-            source
+        // Signal burn-in for image-based subtitle codecs the client does not
+        // declare support for in its device profile. Text codecs are handled via
+        // VTT conversion in the DeliveryUrl loop below.
+        if let Some(idx) = effective_sub_idx {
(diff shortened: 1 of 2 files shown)


Bobbls:
just to be transparent, ASS sub support works, but not perfectly on jellyfin-web, I can't get it to a state where the subtitles always get populated by jellyfin-web without implementing a precache.

Bobbls (commented):
All issues addressed:
- CRLF handling: added `replace("\r\n", "\n")` in both srt_to_vtt and srt_to_jellyfin_json
- PGS muxer: mapped `pgssub` → `sup` for ffmpeg `-f` flag
- Binary CORS: added `Cache-Control` and `Access-Control-Allow-Origin` headers to binary response path
- Empty cache: returns 200 with empty body instead of 500
- is_ass: removed unused variable
- Binary passthrough: separate raw byte path for SUP/PGS
- Non-UTF-8: use `from_utf8_lossy` for cached reads

Issue #14, opened 2026-06-14
Next Up section is not populating in Infuse


raulXgarcia:
The "Next Up" section for Remux in the Infuse interface is currently blank or failing to populate with upcoming titles, even though there is active watch history and next-in-sequence episodes available. The Jellyfin web player shows titles in the "Continue Watching" section. 

To clarify, I'm referring to the "Next Up" catalog list for Remux, not the default Infuse "Up Next List".

PR #21, opened 2026-06-17
fix: respect enable_user_data and normalize NextUp cutoff handling

### LINKED ISSUE #19: Recently watched items do not show up to the left in next up and continue watching
The last thing I watched doesn't end up to the left in the "continue watching" and "next up" queues. It would be nice if they would (left to right) be the order I watched them (most recent to least recent). Right now items just end up somewhere in the queue.

Using: 0.0.1-nightly.20260616.f7b57ab.

Bobbls:
NextUp now orders series by the most recent `last_played_at`, and `enable_user_data` defaults are respected in the affected item paths.

`nextUpDateCutoff` is now normalized before querying so RFC3339 and date-only inputs compare correctly against SQLite datetime values without adding per-row SQL conversion cost.

- default `enable_user_data` to `true` in the affected item/user-data paths
- order NextUp series by `MAX(last_played_at) DESC`
- normalize `nextUpDateCutoff` to SQLite datetime format before binding
- return `400 Bad Request` for invalid cutoff formats
- add focused tests for ordering, RFC3339 cutoff handling, and invalid cutoff rejection

This should resolve #19.

Also bumps `tikv-jemallocator` to restore builds on GCC 16, where the older bundled jemalloc source no longer compiles cleanly in this environment.

### DIFF
--- a/crates/remux-server/src/api/shows.rs
+++ b/crates/remux-server/src/api/shows.rs
@@ -6,12 +6,10 @@ use axum::{
     response::IntoResponse,
 };
 use axum_extra::extract::Query;
-use chrono::{DateTime, NaiveDate, NaiveDateTime};
-use http::StatusCode;
 use remux_macros::{api_query, get};
 use uuid::Uuid;
 
-use crate::{AppState, OptionExt, ResultExt, api, db, db::auth};
+use crate::{AppState, OptionExt, api, db, db::auth};
 use axum_anyhow::ApiResult as Result;
 
 use super::items::get_items;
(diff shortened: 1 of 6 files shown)


Bobbls:
also looking at other open issues this MAY fix #14

crates/remux-server/src/api/shows.rs
```diff
@@ -30,6 +31,37 @@ pub fn livetv_view_item() -> api::BaseItemDto {
     }
 }
 
+fn normalize_next_up_cutoff(cutoff: Option<&str>) -> Result<String> {
```
lostb1t: Add parsing on the struct/field level with serde. Might already be helpers. not sure

PR #24, opened 2026-06-19
fix: prevent panics and surface silent DB errors

OlegA-X:
## Summary

Three independent robustness fixes — each prevents a failure mode where the server either panics (crashing the worker thread) or silently drops errors (causing data inconsistencies).

## 1. Validate `segment_length` from client requests

**File:** `api/playback.rs`

The `SegmentLength` query parameter is user-controlled (`Option<i32>`). It was stored directly into the transcode session without validation:

```rust
let segment_length = q.segment_length.unwrap_or(6) as u32;  // no validation
```

A `SegmentLength=0` causes **divide-by-zero panics** downstream:
- `transcode/engine.rs:193`: `cutoff_idx = (playback_secs - SEGMENT_KEEP_SECS) / segment_length`
- `api/playback.rs:4212`: `segment_gap_threshold = 24 / segment_length`

Negative values (`-1`) silently cast to huge `u32` via `as`.

**Fix:** clamp to `1..=30` at session creation. Jellyfin clients send 6–10; 30 is a generous ceiling.

## 2. Return errors instead of `panic!()` for unsupported stream variants

**File:** `stream.rs` (`StreamDescriptor::into_source`)

`into_source()` panicked for `Rtsp` and `Opendal` descriptors:

```rust
Self::Rtsp { .. } => panic!("Rtsp descriptors must be served through the transcode path"),
Self::Opendal { .. } => panic!("Opendal descriptors must be served through their addon"),
```

If a descriptor of these types reached the direct-stream path (e.g. from addon config or DB state), the worker thread crashed — a crash-per-request DoS.

**Fix:** changed return type to `anyhow::Result<Box<dyn StreamSource>>`, using `anyhow::bail!` for the unsupported variants. The request now fails gracefully with a 500. Updated all 3 call sites (`api/playback.rs` ×2, `api/stream.rs` ×1) to propagate with `?`.

## 3. Log DB-write errors instead of silently dropping them

**Files:** `addons/mod.rs`, `services/resolve.rs`

Three places used `.ok()` to silently discard DB-write errors:

- `addons/mod.rs` — `MediaRelation::delete_by_left_ids(...).await.ok()` before upserting relations
- `services/resolve.rs` — `Media::upsert(...).await.ok()` after resolving media metadata
- `services/resolve.rs` — `persist_from_store(...).await.ok()` in the concurrent-wait path

If these writes failed (DB error, disk full, etc.), the caller continued as if they succeeded, causing **silent inconsistencies** between the in-memory addon cache and the database — stale relations, missing metadata, duplicated entries on next refresh.

**Fix:** replaced `.ok()` with `warn!(error = %e, ...)` so failures are at least visible in logs and debuggable. Not propagated as hard errors (to preserve existing graceful-degradation behavior), but no longer silent.

## Verification

- Individual unit tests pass ✅ (e.g. `test_get_sessions_empty`, `test_playback_start`)
- Note: running the full test suite in parallel shows pre-existing test-isolation failures (62 tests fail on clean `upstream/main` too — shared SQLite/filesystem contention). These are not introduced by this PR.

cc @lostb1t

### DIFF
--- a/crates/remux-server/src/addons/mod.rs
+++ b/crates/remux-server/src/addons/mod.rs
@@ -189,9 +189,9 @@ pub(crate) async fn save_pending_relations(ctx: &AppContext, items: &[db::Media]
         // Don't link relations that point to name-keyed person stubs.
         .filter(|r| !name_keyed_person_ids.contains(&r.right_media_id))
         .collect();
-    db::MediaRelation::delete_by_left_ids(&ctx.db, &all_ids)
-        .await
-        .ok();
+    if let Err(e) = db::MediaRelation::delete_by_left_ids(&ctx.db, &all_ids).await {
+        warn!(error = %e, "failed to delete stale relations before upsert");
+    }
     if !all_rels.is_empty() {
         if let Err(e) = db::MediaRelation::upsert(&ctx.db, &all_rels).await {
             warn!(error = %e, "failed to upsert relations batch");
--- a/crates/remux-server/src/api/playback.rs
+++ b/crates/remux-server/src/api/playback.rs
@@ -1227,7 +1227,7 @@ async fn videos_stream_inner(
         } else {
             descriptor
                 .clone()
-                .into_source()
+                .into_source()?
                 .serve(&state, &headers)
                 .await?
         };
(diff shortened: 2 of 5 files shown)


lostb1t:

PR #44, opened 2026-06-22
fix(subtitles): honor codec aliases in playback decisions

Bobbls:
Some subtitle capability checks were still treating common codec aliases as different formats.

  In practice that meant profiles advertising support for formats like pgs could still fail against probe data using codec
  names like hdmv_pgs_subtitle. The profile layer could accept the subtitle family, while playbackinfo still added
  SubtitleCodecNotSupported for the selected stream.

  This makes subtitle alias handling consistent across the playback decision path.

  - normalize subtitle codec matching in device_profile.rs
  - use the same alias-aware matching in the image-subtitle support check in api/playback.rs
  - add focused coverage for the selected-subtitle playbackinfo case

  Before this change, a focused repro with a selected hdmv_pgs_subtitle stream and a profile advertising pgs support still
  failed with SubtitleCodecNotSupported.

  After this change, the same repro passes and the subtitle stream is delivered through the expected external SUP path instead
  of being rejected as unsupported.

 Copilot loves tests, so tests included.
  AI was definitely used.

### DIFF
--- a/crates/remux-server/src/api/playback.rs
+++ b/crates/remux-server/src/api/playback.rs
@@ -35,7 +35,7 @@ use crate::{
 
 use crate::{
     IntoApiError, OptionExt, ResultExt,
-    device_profile::DeviceProfileExt,
+    device_profile::{DeviceProfileExt, subtitle_codec_matches_profile},
     sdks,
     services::MediaResolveService,
     torrent,
@@ -438,7 +438,7 @@ async fn items_playbackinfo_inner(
                                         s.codec
                                             .as_deref()
                                             .map_or(false, |c| {
-                                                f.eq_ignore_ascii_case(c)
+                                                subtitle_codec_matches_profile(c, f)
                                             })
                                     })
                             })
@@ -732,7 +732,7 @@ async fn items_playbackinfo_inner(
                                 p.format
                                     .as_deref()
                             })
-                            .any(|f| f.eq_ignore_ascii_case(fmt))
+                            .any(|f| subtitle_codec_matches_profile(fmt, f))
                     })
                     .unwrap_or(false)
             };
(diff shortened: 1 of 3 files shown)


crates/remux-server/src/api/playback.rs
```diff
@@ -2152,6 +2132,109 @@ mod tests {
...
+        let (server, guard, token) = authenticated_server().await;
+        let auth = auth_header_with_token(&token);
+        let now = chrono::Utc::now().naive_utc();
+
+        let mut media = crate::db::Media {
+            title: "PGS Alias Test".to_string(),
+            kind: crate::db::MediaKind::Stream,
+            stream_info: Some(crate::stream::StreamInfo {
+                descriptor: crate::stream::StreamDescriptor::http(
+                    "https://test-videos.co.uk/vids/bigbuckbunny/mp4/h264/1080/Big_Buck_Bunny_1080_10s_5MB.mp4"
```
lostb1t: rather not rely on external network calls. Mock these.

crates/remux-server/src/device_profile.rs
```diff
@@ -12,6 +12,33 @@ pub trait DeviceProfileExt {
     fn check_direct_play(&self, media_source: &MediaSourceInfo) -> TranscodeReasons;
 }
 
+pub(crate) fn subtitle_codec_matches_profile(
```
lostb1t, line 41: when working directly with data strings i prefer to use enums where possible. This could be repkaced with an enum and strum
aliases.

example

```rust

#[derive(Debug, Clone, PartialEq, Eq, strum_macros::EnumString, strum_macros::Display)]
#[strum(ascii_case_insensitive)]
pub(crate) enum SubtitleCodec {
    #[strum(serialize = "pgs", serialize = "pgssub", serialize = "hdmv_pgs_subtitle", serialize = "sup")]
    Pgs,
    #[strum(serialize = "subrip", serialize = "srt")]
    Srt,
    #[strum(serialize = "dvd_subtitle", serialize = "dvdsub")]
    DvdSub,
    #[strum(serialize = "dvb_subtitle", serialize = "dvbsub")]
    DvbSub,
    #[strum(serialize = "ass", serialize = "ssa")]
    Ass,
    #[strum(serialize = "webvtt", serialize = "vtt")]
    WebVtt,
    #[strum(serialize = "mov_text", serialize = "tx3g")]
    MovText,
}

pub(crate) fn subtitle_codec_matches_profile(codec: &str, profile_format: &str) -> bool {
    match (
        codec.trim().parse::<SubtitleCodec>(),
        profile_format.trim().parse::<SubtitleCodec>(),
    ) {
        (Ok(a), Ok(b)) => a == b,
        _ => codec.trim().eq_ignore_ascii_case(profile_format.trim()),
    }
}

```

lostb1t (changes requested):

PR #64, opened 2026-06-27
fix: import local TV when source advertises an Episode catalog

jackgdll:
local addon with a content type of  "TV Episodes" scans cleanly (`scan complete … upserted=N`, no errors) and `opendal_files` fills with correct `imdb_id`, `season`, and `episode` rows. 
but the series library stays `0-0 of 0`, `media` table gets no `series`/`season`/`episode` rows. 

the movies `opendal-local` addon works fine.

this is because `catalogs_for_kinds` pre-filters addons with `supports_type`, but the per-catalog post-filter used a raw `kinds.contains()`. A local `"episode"` source advertises a `MediaKind::Episode` catalog, so it was dropped when `RefreshLibraryTask` requested `LIBRARY_KINDS` (Movie, Series, ...), silently skipping the entire local TV library. 

fixed by using `kind_in_type_list` in the post-filter so it matches the pre-filter. 

tested with my media library and local tv shows now show up and are playable.

I used Claude to debug this and find a fix. It wrote the code and tests since I'm not proficient in Rust but I reviewed the logic and it seems sound and cleaned it up to try to match your coding style.

### DIFF
--- a/crates/remux-server/src/addons/mod.rs
+++ b/crates/remux-server/src/addons/mod.rs
@@ -1220,7 +1220,7 @@ impl AddonService {
                     .filter(|c| {
                         c.media_kind
                             .as_ref()
-                            .is_some_and(|mk| kinds.contains(mk))
+                            .is_some_and(|mk| kind_in_type_list(mk, kinds))
                     })
                     .collect(),
                 Err(e) => {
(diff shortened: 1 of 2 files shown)


jackgdll:
@lostb1t oops I had this fix ready locally since yesterday but it looks like you beat me to it with #61  
I forgot to check new PRs before submitting this one, sorry. I can close this one if you want

PR #72, opened 2026-06-29
fix: recompute Season/Episode stable UUID after meta enrichment

jackgdll:
**Issue**
-  `skipping media item with invalid UUID ... UUID mismatch` errors logged after library refresh 
- seasons/episodes mentioned in the errors never land in the DB, whole seasons of specials vanish.
- skipped item's `id` encodes a season one higher than the stored `parent_idx` (same IMDB anchor, episode number unchanged)

**Cause**
- `AddonRuntime::refresh_meta` re-derives the canonical UUID for `Movie`/`Series` only, not for `Season`/`Episode`.
- The recompute block's comment says: "Catalog stubs arrive with a TMDB-keyed UUID; validate() expects the canonical one", it just was never extended to seasons/episodes.

**Fix**
- `addons/mod.rs`: extend the UUID recompute to `Movie | Series | Season | Episode`, deriving the key via `media.media_id_raw()`, guarded on `canonical().is_some()`.
- `db/media.rs`: regression tests for episode and series ids

> I used Claude to debug and fix this then manually reviewed it and wrote the PR description myself

### DIFF
--- a/crates/remux-server/src/addons/mod.rs
+++ b/crates/remux-server/src/addons/mod.rs
@@ -1416,22 +1416,22 @@ impl AddonService {
             }
         }
 
-        // Recompute stable UUID for Movie/Series once the canonical external ID
+        // Recompute stable UUID for Movie/Series/Season/Episode once the canonical external ID
         // (IMDB or custom stremio) is resolved by meta enrichment. Catalog stubs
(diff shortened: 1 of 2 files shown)


jackgdll:
i'm not sure whether `Person` should also be supported here or not?

lostb1t:
people are handled in a different location so this looks good

PR #84, opened 2026-07-03
dark theme for dashboard

martysl:
change theme to dark for dashboard so eyes not hurt:)


lostb1t:
should be at minimum an option to choose between light and dark theme in the dashboard.

PR #97, opened 2026-07-13
feat: show users on login screens

BrandonItaly:
makes users visible on login screens with the option to hide them

image

### DIFF
--- a/crates/remux-server/src/api/startup.rs
+++ b/crates/remux-server/src/api/startup.rs
@@ -105,6 +105,9 @@ pub async fn post_startup_user(
             None,
         )?;
         user.is_admin = true;
+        let mut policy = crate::api::UserPolicy::default();
+        policy.is_hidden = true;
+        user.policy = Some(sqlx::types::Json(policy));
         user.save_by_username(
             &state
                 .ctx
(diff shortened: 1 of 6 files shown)


crates/remux-server/src/api/startup.rs
```diff
@@ -105,6 +105,9 @@ pub async fn post_startup_user(
             None,
         )?;
         user.is_admin = true;
+        let mut policy = crate::api::UserPolicy::default();
+        policy.is_hidden = true;
```
lostb1t, line 109: is hidden is already true by default.

think we should make policy non optional for a user (struct) so its already there.

crates/remux-server/src/api/users.rs
```diff
@@ -761,8 +761,32 @@ pub async fn update_user(
...
+    let dtos: Vec<api::UserDto> = all_users
+        .into_iter()
+        .filter(|u| {
+            let policy = u.policy.as_ref().map(|p| &p.0);
+            let is_hidden = policy.map(|p| p.is_hidden).unwrap_or(true);
+            !is_hidden
+        })
+        .map(|u| {
+            let mut dto = api::db_user_to_dto(&state.ctx.config.data_dir, u);
+            dto.policy = api::UserPolicy {
```
lostb1t, line 780: the same thing is reassigned here for no reason or is it to hide sensitive stuff for public?

BrandonItaly: it's to avoid sending per-user policy/config over the public route
https://github.com/lostb1t/remux/pull/97#discussion_r3569648725

PR #105, opened 2026-07-14
fix(stream_groups): keep distinct MediaSource IDs on item detail

### LINKED ISSUE #104: Wrong version played / displayed while playing
When I am on an episode I see all versions, but when I start playing the episode I see only one version on the app and the app plays always the same version, I don’t know why. 

Senplayer / Remux nightly.

Image

Image

squelix:
Revert the all-sources item-id stamp from 3ed7594 back to first-only. Stamping every MediaSource with the parent item id collapsed all versions to one MediaSourceId, so the client could only ever request the item id and select_streams always truncated to the first stream (issue #104). Only the first source now takes the item id; the rest keep their distinct group/stream UUID, matching the PlaybackInfo endpoint.

Fixes #104

### DIFF
--- a/crates/remux-server/src/api/models.rs
+++ b/crates/remux-server/src/api/models.rs
@@ -940,11 +940,13 @@ pub fn db_media_to_item(media: db::Media, hide_sources: bool) -> BaseItemDto {
                     .into_iter()
                     .map(MediaSourceInfo::from)
                     .collect();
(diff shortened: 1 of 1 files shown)


squelix:
Fixed already 4a98576b31335e66d5e37299e90d37e38eb2ae09

PR #128, opened 2026-07-21
feat(sessions): admin session revocation and activity log

### LINKED ISSUE #11: Add UI Controls for Session Logout / Token Revocation
Expose session/device logout controls inside the  admin interface.

- Allow users to log out signed-in devices/sessions. (Not required until there may be a user portal?)
- Allow admins to revoke a user’s active sessions.
- On password change, force all other active sessions to re-authenticate.
- Show device/session details where Jellyfin exposes them.
- Add audit logging for admin-triggered logouts.

Jellyfin already supports user password changes and can immediately kick users when disabled, so this should first investigate existing Jellyfin session/auth APIs before building custom session logic.

sean-wils:
Closes #11


sean-wils:
## Live test results

Tested on homelab (Proxmox LXC, Docker, jellyfin-ffmpeg). All scenarios below ran against the deployed branch.

---

### Device list — multi-user sessions with \"This Session\" badge

All connected devices shown across users, with real usernames (not UUIDs), client app, and last-seen time. The current session is identified server-side and flagged with a badge — no token is sent to the client.

image

---

### Single device revoke — confirmation dialog

Screenshot 2026-07-22 at 18 22 39

Confirmed: after confirming, the device disappears from the list immediately.

---

### Revoke All for a user — caller's own session is preserved

\"Revoke All for wilsse\" was triggered from the admin's own session. After confirming, every other wilsse session was removed. The admin's own Browser session (marked \"This Session\") remained — the server explicitly excludes the caller's token from bulk deletion.

Screenshot 2026-07-22 at 18 16 30

---

### Revoked session redirects to login

After a session is revoked, the next API call from that tab returns 401. The dashboard detects this, clears credentials from localStorage, and renders the login screen — no broken \"unauthorized\" state left behind.

Screenshot 2026-07-22 at 18 23 03

---

### Session revoke from Edit User

Sessions can also be revoked inline from the user edit form under Access → Users, without needing to go to the dashboard.

Screenshot 2026-07-22 at 18 30 12

---

### Password change invalidates other sessions

Changed password for wilsse. All other wilsse sessions (Firefox Android, second browser) were immediately invalidated. The session that performed the password change survived.

Screenshot 2026-07-22 at 18 31 04

---

### Admin activity log

Every action is recorded with full date+time, actor username, action type, target user, and device name. Format is \"DD Mon HH:MM\" — readable at a glance even weeks later.

Screenshot 2026-07-22 at 18 23 27

sean-wils:
Taking out of draft. I've reviewed and tested this in earnest; providing screenshots where I can of the live UI tests I performed on my home server.

lostb1t:
i dont think the actual device management should happen on the dashboard. Dashboard is mostly just metrics. move to a separate page

sean-wils:
Sure thing. Let me see what I can do.

sean-wils:
## Live UI Testing: Device Revocation Scenarios

Tested with three simultaneous browser sessions across two users (`wilsse` (user1) + `user2`). All scenarios verified end-to-end.

---

### Setup

The new `/devices` page (under the **Devices** sidebar group) lists all active sessions with individual **REVOKE** buttons and per-user **REVOKE ALL FOR {USER}** bulk actions. The dashboard home shows a read-only active device count — no revoke buttons.

> image — Devices page with 3 active sessions (all wilsse), "REVOKE ALL FOR WILSSE" visible; dashboard showing 2 active devices in the read-only card

---

### Scenario 1 — Revoke a single device (same user)

Clicking **REVOKE** on any device shows a confirmation dialog before acting.

> image — "Revoke this device? It will be signed out immediately." confirmation dialog

After confirming, the revoked session redirects to the login page on its next API call. Remaining sessions are unaffected.

> image — Revoked browser on login page (top right), 2 remaining devices still active (bottom right)

---

### Scenario 2 — Revoke a single device (cross-user)

Works identically across users. An admin can revoke any individual device regardless of which user owns it.

> image — Three-browser setup: Dashboard (left, 2 wilsse + 1 user2), Devices page (top right, both "REVOKE ALL" buttons visible), Activity log (bottom right)
> image — Single-device revoke confirmation with 2-user setup visible
> image — After revoking user2's first device: revoked browser on login, user2's remaining "THIS SESSION" device still present

---

### Scenario 3 — Revoke all devices for another user

The **REVOKE ALL FOR {USER}** bulk action triggers its own confirmation dialog.

> image — "Revoke ALL devices for this user? They will be signed out everywhere." confirmation dialog

After confirming, all sessions for the target user are terminated. The admin's own session (different user) is completely unaffected.

> image — Activity log showing "All sessions revoked user:user2" (top left), revoked browser on login (top right), wilsse "THIS SESSION" + user2 "THIS SESSION" both still active on Devices page (bottom right — user2's own device survives because they were the one performing the revoke)

---

### Scenario 4 — Self-revocation: current session is preserved

When a user revokes all devices for their *own* account, the session that initiated the revoke is preserved. All other sessions for that user are terminated.

> image — Admin (user2, THIS SESSION) confirming "Revoke ALL devices for wilsse"; two other wilsse browsers visible in the device list
> image — Result: both wilsse browsers redirected to login, user2's session untouched with only their "THIS SESSION" device remaining

This matches Jellyfin's behaviour: `delete_all_for_user` skips the calling session's `access_token`, so the current session's token is never invalidated.

---

### Audit trail

Every revocation event — single or bulk — is written to the activity log immediately, with actor, target user, device name, and timestamp.

> Screenshot 2026-07-24 at 23 16 13 — Activity log showing "Session revoked" and "All sessions revoked" entries with full attribution

The full revocation history across all test scenarios:

> image (left panel) — Complete activity log spanning all test runs

---

### Revocation mechanism

No heartbeat or push notification is used (matching Jellyfin's approach). The revoked session hits a 401 on its next API call, which triggers `AppState::execute()` → `logout()` → redirect to login. In practice this means the redirect happens on the next user interaction or background refresh.

sean-wils:
@lostb1t it's probably worth getting Copilot to give it a couple of passes.

lostb1t:
@sean-wils think it has merge conflicts. if you can fix it i will start testing this.

sean-wils:
Hey @lostb1t I'm not seeing any merge conflicts against [lostb1t:main](https://github.com/lostb1t/remux/tree/main).
Should I be pointing the source of this PR elsewhere?

lostb1t:
did you try an rebase from main?

sean-wils:
Ahh gotcha; sorry I was relying on what I could see from here. Pushing now.

lostb1t:
yeah need to figure out howto retrigger a pr check after main changes

anyway, getting some sql errors

```

grations. Do not interrupt!
Error: while executing migration 202607300001: error
 returned from database: (code: 1) Cannot add a colu
mn with non-constant default

Caused by:
    0: error returned from database: (code: 1) Canno
t add a column with non-constant default
    1: (code: 1) Cannot add a column with non-consta
nt default

```

sean-wils:
@lostb1t give it a try now.

lostb1t:
some servers have thousands of users. so needs pagination. Both devices and actvity.

same for the revoke all buttons. that will not work for a lot of users. And dont think those actions should be at the bottom. shiuld
be at the top

styling on mobile is off

image

sean-wils:
All good points. I can take a look.

sean-wils:
Screenshot 2026-08-05 at 02 23 51

lostb1t:
looks better!

Im going to nitpick one last time. can we keep pagination styled the same? we already have pagination on iptv for example. dont care which one as long as they are the same

image

sean-wils:
> we already have pagination on iptv for example. dont care which one as long as they are the same
> 

Ahh nice! Glad there's something I can reference, I was going nuts trying to come up with something.

I'll take another look. 👍

Issue #131, opened 2026-07-22
Bug with fankai addon


squelix:
I tried to add this addon in stremio, and it's working, but in remux its not. Maybe its a bug ?

https://streamio.fankai.fr/configure

<img width="1552" height="230" alt="Image" src="https://github.com/user-attachments/assets/d87ff2c4-b8a4-4c13-90b1-7ef6f98f6d02" />

<img width="1849" height="915" alt="Image" src="https://github.com/user-attachments/assets/f91b48c2-5c94-409b-a1ba-c6567a42b855" />

<img width="665" height="602" alt="Image" src="https://github.com/user-attachments/assets/90562b4a-fda7-4392-b17b-400ff287a985" />

<img width="1857" height="894" alt="Image" src="https://github.com/user-attachments/assets/21957eaa-0d36-4427-8add-907278b727dc" />

PR #133, opened 2026-07-22
fix: support Stremio addons with non-standard content types (fankai/anime)

squelix:
## Summary

Fixes the fankai (FKStream) addon catalog end-to-end. The catalog import
crashed, then paginated forever, then imported items were invisible in
collections, then had no seasons/episodes, then had no playable streams —
each layer was masked by the next, so this PR fixes all of them together.

### 1. Catalog import crash
- `Meta.genre` / `Meta.genres` now accept a bare string (some addons send a
  single string instead of an array), reusing the existing tolerant
  deserializer already used for `country`/`director`/`cast`/`writer`.
- `Meta.runtime` now treats an unparseable value (e.g. fankai's `"En cours"`
  status string, not a duration) as absent instead of failing the whole
  catalog page's deserialization.

### 2. Infinite pagination
`get_catalog_stream` continued paginating on *any* fetch error instead of
stopping — once the addon's real last page was reached and it started
returning 404 instead of an empty array, this hammered the addon with
hundreds of requests (`skip=79` through `skip=3397+`). Now a fetch error
stops pagination (expected/normal terminal condition, downgraded to `debug!`).

### 3. Imported items invisible in collections
The release-date filter (`FilterByDigitalReleaseDate`) hid every item with
no resolvable release date at all — some addon catalogs (fankai included)
never report one. Movies/series with no known date now pass the filter
(unknown ≠ future release); seasons/episodes keep the strict "hide if
undated" behavior since a missing air date there really does mean unaired.

### 4. No seasons/episodes
`fetch_and_cache_meta` derived the Stremio type to query from `MediaKind`
(→ `"series"`), 404ing against addons using a custom type (fankai's
`"anime"`). Fixed by:
- Tracking the addon's own non-standard type string via
  `ExternalIds::custom_stremio_type`, propagated onto seasons/episodes, and
  preferring it over the generic derived type.
- When a custom-ID item still 404s (e.g. rows created before this field
  existed), consulting the addon's own manifest for its declared type(s) and
  retrying instead of failing permanently (`manifest_meta_type_fallback`).

### 5. No playable streams
Two separate bugs, both silent (no error, just an empty stream list):
- The stream lookup reconstructed the video id as
  `"{series}:{season}:{episode}"` — only a convention for series-type
  addons, not a guarantee. Now the addon's own video id
  (`videos[].id`, captured as `ExternalIds::custom_stremio_id`) is used
  directly, falling back to reconstruction for episodes stored before this
  field was captured.
- `supports_type()` collapsed any unrecognized manifest type (e.g. `"anime"`)
  to `MediaKind::Movie`, which made the addon registry believe an
  anime/series-only addon served *only* movies — silently excluding it from
  `addons_for::<dyn StreamAddon>` for every Series/Season/Episode lookup, so
  no stream addon was ever even attempted for episodes. Unrecognized types
  are now left out of `supported_types` entirely, which correctly bypasses
  the manifest-type gate instead of asserting a wrong one.

## Test plan
- [x] `cargo test -p remux-server --lib -- --test-threads=1` — 251 passed, 0 failed
- [x] `cargo fmt --all -- --check` clean
- [x] Verified end-to-end against the real fankai addon: catalog import,
      collection visibility, season/episode tree, and stream playback for
      "Bleach Yabai" S1E1 all confirmed working via live DB inspection and
      manual playback test

### DIFF
--- a/crates/remux-sdks/src/stremio/mod.rs
+++ b/crates/remux-sdks/src/stremio/mod.rs
@@ -408,6 +408,7 @@ pub struct Meta {
     )]
     pub writer: Option<Vec<String>>,
     pub description: Option<String>,
+    #[serde(default, deserialize_with = "deserialize_option_string_or_array")]
     pub genre: Option<Vec<String>>,
     #[serde(
         default,
@@ -440,6 +441,7 @@ pub struct Meta {
     )]
     pub popularity: Option<f64>,
     pub id: String,
+    #[serde(default, deserialize_with = "deserialize_option_string_or_array")]
     pub genres: Option<Vec<String>>,
     // pub season_posters: Option<Vec<String>>,
     // this can be a range 2012-2015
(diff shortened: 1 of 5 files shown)


squelix:
Fix #131

lostb1t:
does not adhere to our AI policy

PR #136, opened 2026-07-23
fix: Jellyfin API compat — Filters2 endpoint and list filter params

OlegA-X:
Connected the Swiparr client to remux and noticed that applying filters (by year, genre, etc.) caused everything to crash with an error. Dug into it and found two issues, both related to compatibility with the public Jellyfin API.

1. GET /Items/Filters2 → 400. Clients send paths in TitleCase, while remux routes use lowercase. Case conversion was skipping segments containing digits, so Filters2 was not being converted to filters2. As a result, the route fell back to /items/{id} and attempted to parse the word "Filters2" as a UUID. Now, purely alphabetical segments along with known keywords containing digits (like Filters2 and Hls1) are converted to lowercase. UUIDs and device/session IDs remain untouched.

2. GET ?Years=2020,2021 → 400. The years, genres, official_ratings, and tags fields were unable to parse string values into arrays. Added deserializers: comma-separated for years, and pipe-separated (|) for genres, ratings, and tags (matching Jellyfin's implementation).

Tested on a live setup with remux + Swiparr: filters are working properly, and basic requests haven't broken. 
And added 13 unit tests.

### DIFF
--- a/crates/remux-sdks/src/remux/mod.rs
+++ b/crates/remux-sdks/src/remux/mod.rs
@@ -1523,7 +1523,9 @@ where
 /// Jellyfin uses `|` for multi-value genre/rating/tag filters (e.g.
 /// `Genres=Action|Comedy`, `OfficialRatings=PG|R`). We also accept commas so the
 /// field round-trips through our own `serialize_comma_opt` serializer.
-fn deserialize_pipe_comma_str<'de, D>(deserializer: D) -> Result<Option<Vec<String>>, D::Error>
+fn deserialize_pipe_comma_str<'de, D>(
+    deserializer: D,
+) -> Result<Option<Vec<String>>, D::Error>
 where
     D: Deserializer<'de>,
 {
@@ -1562,7 +1564,9 @@ where
 }
 
 /// `OfficialRatings=PG|R` → `vec!["PG", "R"]`.
-pub fn deserialize_official_ratings<'de, D>(d: D) -> Result<Option<Vec<String>>, D::Error>
+pub fn deserialize_official_ratings<'de, D>(
+    d: D,
+) -> Result<Option<Vec<String>>, D::Error>
 where
     D: Deserializer<'de>,
 {
@@ -1687,13 +1691,19 @@ mod tests {
     #[test]
     fn genres_filter_parses_pipe_separated() {
         let q = parse_query("Genres=Action|Comedy");
-        assert_eq!(q.genres, Some(vec!["Action".to_string(), "Comedy".to_string()]));
+        assert_eq!(
+            q.genres,
+            Some(vec!["Action".to_string(), "Comedy".to_string()])
+        );
     }
 
     #[test]
     fn genres_filter_also_accepts_comma_separated() {
         let q = parse_query("Genres=Action,Comedy");
-        assert_eq!(q.genres, Some(vec!["Action".to_string(), "Comedy".to_string()]));
+        assert_eq!(
+            q.genres,
+            Some(vec!["Action".to_string(), "Comedy".to_string()])
+        );
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


crates/remux-server/src/lib.rs
```diff
@@ -541,16 +541,22 @@ pub fn rewrite_request_uri<B>(mut req: http::Request<B>) -> http::Request<B> {
...
-    // Smart-lowercase: only lowercase purely-alphabetic segments (route
-    // keywords like "Sessions", "Playing"). Path parameter values — UUIDs,
-    // base64 device IDs — contain digits and are preserved unchanged.
+    // Smart-lowercase: lowercase route-keyword segments (Sessions, Items, …) but
+    // leave path-parameter values (UUIDs, base64 device/session ids, …) untouched,
+    // since they are looked up case-sensitively. A segment is lowercased when it
+    // is purely alphabetic, or when it matches a known keyword that contains digits
+    // (Filters2, Hls1).
+    static DIGIT_KEYWORDS: &[&str] = &["filters2", "hls1"];
     let smart_lower_path: String = path
```
lostb1t, line 544: uuids and device ids are handled properly since i introduced this method. we can just check against is_ascii_alphanumeric so drop the number list

crates/remux-sdks/src/remux/mod.rs
```diff
@@ -1501,6 +1511,84 @@ where
...
+    D: Deserializer<'de>,
+{
+    deserialize_comma_str(d)
+}
+
+/// Deserialize string filters that Jellyfin clients send pipe-separated.
+/// Jellyfin uses `|` for multi-value genre/rating/tag filters (e.g.
+/// `Genres=Action|Comedy`, `OfficialRatings=PG|R`). We also accept commas so the
+/// field round-trips through our own `serialize_comma_opt` serializer.
+fn deserialize_pipe_comma_str<'de, D>(
```
lostb1t: lets just replace all deserialize_comma_str occurrences
in the remux api layer with this.

i dont see an need to be strict about comma-only anywhere

lostb1t (changes requested):

crates/remux-sdks/src/remux/mod.rs
```diff
@@ -1501,6 +1511,84 @@ where
...
+            .iter()
+            .flat_map(|s| split_one(s))
+            .collect(),
+        None => return Ok(None),
+    };
+    Ok(Some(values))
+}
+
+/// `Genres=Action|Comedy` → `vec!["Action", "Comedy"]`.
+pub fn deserialize_genres<'de, D>(d: D) -> Result<Option<Vec<String>>, D::Error>
```
lostb1t: all these wrappers are not needed.

use it directly 

```

#[serde(deserialize_with = "deserialize_pipe_comma_str", ...)]
pub years: Option<Vec<i64>>,

#[serde(deserialize_with = "deserialize_pipe_comma_str", ...)]
pub genres: Option<Vec<String>>,

```

lostb1t (changes requested):

crates/remux-server/src/lib.rs
OlegA-X: On dropping the keyword list in favor of bare is_ascii_alphanumeric - I think this regresses /Sessions/{sessionid}, and would rather keep the list unless I'm missing something.

Jellyfin web-client device ids are base64 and can be fully alphanumeric with no hyphens - a real one I observed: TW96aWxsYS81... (128 chars, 0 non-alphanumeric chars).Device::get_by_id matches via case-sensitive WHERE id = ?, and /Sessions/{sessionid} is a {sessionid} route, so lowercasing that segment yields 404 session not found and breaks remote control. I hit exactly this in an earlier iteration of this PR, which is why the explicit keywordlist is there.

If device ids are guaranteed to be percent-encoded in the path (so %2F/%2B make the segment non-alphanumeric and it's skipped), then bare is_ascii_alphanumeric is safe and I'll drop the list. I couldn't find where that's enforced though-could you point me at it?

lostb1t: yeah i sneaky fixed that in main before asking you to change it. pull in main and it shoukd work as it makes the device id queries case insensitive

OlegA-X: Done.Dropped the keyword list, using is_ascii_alphanumeric directly.

But I'm still hitting a 404 on /Sessions/{sessionid}/Command with an alphanumeric device id. Pulled main (verified origin/main is in sync with upstream), but Device::get_by_id is still SELECT * FROM devices WHERE id = ?1 with no COLLATE NOCASE, and the devices.id column has no COLLATE NOCASE either. The rewrite lowercases the id before the lookup, so it doesn't match the stored record.

Logged in with device id TW96aWxsYS81TestDeviceABC123 (shows up in /Sessions), then POST /Sessions/{that-id}/Command → 404 session not found.

Am I missing where the case-insensitive lookup landed?

lostb1t: forgot to push. sorry about that. Merged! tnx!

PR #138, opened 2026-07-23
fix(streams): resolve addon streams on demand in the stream/HLS path

OlegA-X:
Music clients such as Kolis Music, Feishin, and Finamp failed to play tracks fetched from a metadata addon with `no playable sources`, unless the track had been played first through the web UI.

Some clients hit the audio stream endpoints directly, skipping PlaybackInfo. That path (`StreamService::lookup`) only read already-persisted streams and never resolved them, so a fresh addon-backed track had nothing to serve until a PlaybackInfo call warmed it up.

Tested on a live setup: fresh tracks now play on the first request.

### DIFF
--- a/crates/remux-server/src/api/playback.rs
+++ b/crates/remux-server/src/api/playback.rs
@@ -505,6 +505,7 @@ async fn items_playbackinfo_inner(
 pub async fn items_file(
     headers: headers::HeaderMap,
     State(state): State<AppState>,
+    session: auth::AuthSession,
     Path(id): Path<Uuid>,
     Query(mut q): Query<api::VideoStreamQuery>,
 ) -> Result<impl IntoResponse> {
@@ -527,7 +528,7 @@ pub async fn items_file(
     let safe = filename
         .replace('"', "")
         .replace('\\', "");
-    let mut response = videos_stream_inner(headers, state, id, q)
+    let mut response = videos_stream_inner(headers, state, session, id, q)
         .await?
         .into_response();
     if let Ok(val) =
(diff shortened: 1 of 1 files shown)


crates/remux-server/src/services/stream_service.rs
```diff
@@ -212,9 +212,29 @@ impl StreamService {
...
-                let sources = media
+                let mut sources = media
                     .streams(&ctx.db)
                     .await?;
+                // Some clients (e.g. Kolis Music) hit the stream/HLS endpoints
+                // directly, skipping PlaybackInfo. That path never resolves
+                // streams, so a fresh addon-backed track (Deezer metadata + a
+                // separate stream addon) with no cached child stream would fail
+                // with "no playable sources". Resolve on demand here, mirroring
+                // `StreamService::load()`, then re-read.
```
sean-wils: I'm not sure the comments need to be so verbose, if they're at all needed.

crates/remux-server/src/services/stream_service.rs
```diff
@@ -212,6 +212,14 @@ impl StreamService {
             }
             db::MediaKind::Movie | db::MediaKind::Episode | db::MediaKind::Track => {
                 let mut media = media;
+                // Some clients hit the stream endpoints directly, skipping
+                // PlaybackInfo. Resolve addon streams here so fresh tracks play
+                // without a prior PlaybackInfo call. Cheap when fresh (TTL).
+                ctx.addons
+                    .refresh_streams(&mut media, ctx, None)
```
lostb1t: I would move this to the handler thats needs it tho, its not the an service layer concern. And that way, a user can also be passed to it.

lostb1t (changes requested):

lostb1t:
fixed in main

PR #141, opened 2026-07-24
fix: clean tag providers

### LINKED ISSUE #73: Tags display needs cleanup
Image

I suggest the tags need to drop the 'provider:' prefix for a cleaner look

squelix:
Try to fix #73

**Summary**

- TMDB watch-provider tags are stored with a prefix (provider: HBO Max), that's an internal discriminator used by `save_pending_tags` (refresh without overwriting user-set tags) and the provider API filter. That prefix was leaking into the item detail page.
- Hid the prefix on the client side only: the injected JS now cleans up the tag link's label text, the href stays untouched clicking a tag still filters on the raw provider: HBO Max value in the DB.
_No server-side DB/refresh/filter changes_.

**Bonus, if accepted:**

- Moved the injected JS (web_patches.rs) into its own file (assets/web-patches.js, loaded via include_str!) so it can be tested in isolation.
- Added a jsdom test harness (tests/web-patches-js, Node node:test) covering the new tag stripper, plus the existing "Recently Added in " stripper as a bonus (same pattern). Wired into cargo make tests via a new test-js-patches task.


lostb1t:
this needs to be done server side because we cant control all clients.

if you want to discuss it we can on discord

squelix:
> this needs to be done server side because we cant control all clients.
> 
> if you want to discuss it we can on discord

Oh ok i see, i close this MR an retry later ;)

PR #143, opened 2026-07-25
feat: support nested collection browsing and group containers

### LINKED ISSUE #126: feat: collection groups / sub-folders for organising collections by type
Currently all collections live under a flat "Collections" section. It would be useful to be able to group collections into named folders so the browse view doesn't become one long undifferentiated list as the collection count grows.

## Example use cases

**By content type + streaming service**
The same streaming service naturally has both films and series, but you'd want them separated:
- `TV` → Netflix Series / BBC Drama / Anime / Prime Video TV
- `Films` → Netflix Films / A24 / Prime Video Films / Disney+

Without grouping there's no clean way to surface `TV › Netflix` and `Films › Netflix` as separate children — a flat "Netflix" collection mixes both.

**By curation theme**
- `Best Of` → IMDb Top 250 / Criterion Collection / Oscar Winners / Hidden Gems
- `Genres` → Action / Sci-Fi / Horror / Documentary

**By era or mood**
- `Decades` → Best of 70s / Best of 80s / Best of 90s
- `Mood` → Feel-Good / Date Night / Background TV

## Proposed behaviour
- Admin can assign a collection to a named group (e.g. `group: Films`)
- The client browse view shows group folders containing their member collections
- Promoted home screen rows remain unaffected — groups are a browse-layer concept only

## Current workaround
- Name-prefix collections ("TV: ...", "Films: ...") — visually groups in sort order but no actual folder hierarchy
- Use Smart Collections with `media_type` filter promoted as separate home screen rows — works for type splits but not thematic grouping

sean-wils:
POC that could potentially close https://github.com/lostb1t/remux/issues/126

### DIFF
--- a/crates/remux-dashboard/src/pages/collections.rs
+++ b/crates/remux-dashboard/src/pages/collections.rs
@@ -12,14 +12,6 @@ fn is_group_container(item: &BaseItemDto) -> bool {
     item.collection_type
         .as_ref()
         == Some(&CollectionType::Boxsets)
-        && item
-            .remux
-            .as_ref()
-            .and_then(|r| {
-                r.collection_kind
-                    .as_ref()
-            })
-            == Some(&RemuxCollectionKind::Manual)
 }
 
 fn is_promoted(item: &BaseItemDto) -> bool {
@@ -918,8 +910,9 @@ pub fn CollectionForm(
 #[component]
 fn CollectionGroupChildren(app_state: AppState, collection_id: String) -> Element {
     let mut all_collections: Signal<Vec<BaseItemDto>> = use_signal(Vec::new);
-    let mut current_children: Signal<Vec<(String, String)>> = use_signal(Vec::new);
+    let mut current_children: Signal<Vec<String>> = use_signal(Vec::new);
     let mut loading = use_signal(|| true);
+    let mut child_error: Signal<Option<String>> = use_signal(|| None);
 
     let col_id = collection_id.clone();
     let client = app_state
@@ -929,7 +922,7 @@ fn CollectionGroupChildren(app_state: AppState, collection_id: String) -> Elemen
         let col_id = col_id.clone();
         let client = client.clone();
         spawn(async move {
-            let children = client
+            let child_items: Vec<BaseItemDto> = client
                 .execute(GetCollectionItems {
                     collection_id: col_id.clone(),
                     start_index: None,
@@ -937,22 +930,18 @@ fn CollectionGroupChildren(app_state: AppState, collection_id: String) -> Elemen
                 })
                 .await
                 .ok()
-                .map(|r| {
-                    r.items
-                        .into_iter()
-                        .filter_map(|item| {
-                            let media_id = item
-                                .id
-                                .to_string();
-                            let relation_id = item.playlist_item_id?;
-                            Some((media_id, relation_id))
-                        })
-                        .collect::<Vec<_>>()
-                })
+                .map(|r| r.items)
                 .unwrap_or_default();
-            current_children.set(children);
+            let child_ids: Vec<String> = child_items
+                .iter()
+                .map(|item| {
+                    item.id
+                        .to_string()
+                })
+                .collect();
+            current_children.set(child_ids);
 
-            let all = client
+            let mut all: Vec<BaseItemDto> = client
                 .execute(GetItems(GetItemsQuery {
                     include_item_types: Some(vec![MediaType::BoxSet]),
                     include_childless: Some(true),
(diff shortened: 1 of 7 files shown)


lostb1t:
im fine with the idea. Lemme know when it's ready for review.

sean-wils:
## Live UI testing scenarios

### Dashboard

1. **Collection list with badges** 
1.1 shows **LIBRARY** badge on promoted groups, **GROUP** badge on non-promoted groups
   - Screenshot 2026-07-26 at 17 26 02

2. **Edit promoted group (`Films (Test)`)**
2.1  Media Kind = Collections, Promote to Library ON, child picker with selected children
   -  Screenshot 2026-07-26 at 17 26 36

3. **Edit non-promoted group (`TV Series (Test)`)**
3.1. Promote to Library OFF, different children selected
   - Screenshot 2026-07-26 at 17 26 50

4. **Edit root `Collections`**
4.1 child picker excludes promoted libraries, only non-promoted collections available
   - Screenshot 2026-07-26 at 17 27 18
   - Screenshot 2026-07-26 at 17 27 35

### Jellyfin client

5. **Sidebar**
5.1 promoted group `Films (Test)` appears as a library entry in the sidebar
   - Screenshot 2026-07-26 at 17 27 55

6. **Promoted group browse**
6.1 clicking `Films (Test)` shows landscape grid with only its explicit child collections (7 items)
   - Screenshot 2026-07-26 at 17 28 40

7. **Collections index**
7.1 browsing `Collections` shows ungrouped collections + non-promoted groups + pinned overrides, no duplicates of children already assigned to other groups (6 items)
   - Screenshot 2026-07-26 at 17 29 07

8. **Non-promoted group browse**
8.1 clicking `TV Series (Test)` from within `Collections` shows the same landscape grid view as the promoted group (13 items)
   - Screenshot 2026-07-26 at 17 30 33

Tested on `Jellyfin Web 10.10.7`, `Wholphin 1.0.3`, and `Streamyfin 0.54.1`.

sean-wils:
@lostb1t I managed to get this working. Pretty happy with the result.
The screenshots aren't pretty but it demonstrates full functionality.

I put my QA hat on and nailed down a few bugs ahead of time, so I think we're good. But anything you notice please shout.

crates/remux-server/src/db/media.rs
```diff
@@ -3018,6 +3023,35 @@ impl Media {
             if let Some(ref f) = filter.filter_rules {
                 apply_filter_rules(qb, f);
             }
+            if filter.exclude_grouped {
```
lostb1t: nitpick, we also have stream groups so maybe make the argument more explicit like exclude collection groups

crates/remux-dashboard/src/pages/collections.rs
```diff
@@ -857,3 +919,157 @@ pub fn CollectionForm(
...
+                                            r.collection_kind
+                                                .as_ref()
+                                        })
+                                        == Some(&RemuxCollectionKind::Manual))
+                                && !item
+                                    .remux
+                                    .as_ref()
+                                    .and_then(|r| r.promoted)
+                                    .unwrap_or(false)
+                        })
```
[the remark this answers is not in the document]
sean-wils: This is by design. I don't see why you'd want to include/duplicate a promoted collection in a group.

lostb1t:
we have a filtering system for media items. (FilterRule)
 
Wondering why we are not using that on collections kind aswell? would
be only this filter for now by im sure others will
follow like tags

sean-wils:
Nice idea. Let me take a look.

sean-wils:
Done.

- Smart collection groups uses the existing filter functionality.
- Manual collection groups uses the new child selection functionality.

crates/remux-server/migrations/202607300001_device_created_at_and_activity_log.sql
lostb1t, line 1: dont think this file
should be here...

crates/remux-server/src/db/media.rs
```diff
@@ -6673,6 +6718,32 @@ fn filter_rule_to_sql(rule: &remux_sdks::remux::FilterRule) -> Option<(String, b
             Some((sql, negated))
         }
         R::Catalog { .. } => None,
+        R::GroupContainer { value } => {
```
lostb1t, line 6993: when having a lot of collections this will start to have performance issues.

im actually wondering if multi
parent is really needed? if we keep it single parent then we can just use existing parent id flows and make is simpler. 

not saying no just wondering if multi parent is really needed because i dont think so

lostb1t (changes requested):

crates/remux-server/migrations/202607300001_device_created_at_and_activity_log.sql
sean-wils: My bad. I probably forgot to stash changes.

sean-wils: Should no longer be there after rebase.

crates/remux-server/src/db/media.rs
sean-wils: Yeah fair, good call out. I've switched to single-parent `parent_id`. The group container is now just `media.parent_id = group.id` on the child row.

The `GroupContainer` filter rule is just `media.parent_id IS NOT NULL` since these queries are already scoped to collections.

lostb1t:
fixed in 205

PR #145, opened 2026-07-25
feat(stream-groups): add AudioLanguage rule from probe_data

OlegA-X:
Added an AudioLanguage rule to stream groups that reads audio track languages from probe data (populated by RemuxDB), so a group can select e.g. only English-audio sources.

matches() now takes the parent Media's probe_data, since StreamInfo.probe_data is almost always empty. 

### DIFF
--- a/crates/remux-sdks/src/remux/mod.rs
+++ b/crates/remux-sdks/src/remux/mod.rs
@@ -1643,8 +1643,15 @@ mod tests {
         assert_eq!(language_label("rus"), "Russian");
         assert_eq!(language_label("ENG"), "English");
         assert_eq!(language_label("fra"), "French");
-        // Unknown code falls back to the raw string.
+        assert_eq!(language_label("deu"), "German");
         assert_eq!(language_label("xxx"), "xxx");
+    }
+
+    #[test]
+    fn normalize_lang_code_maps_terminologic_to_bibliographic() {
+        assert_eq!(normalize_lang_code("fra"), "fre");
+        assert_eq!(normalize_lang_code("DEU"), "ger");
+        assert_eq!(normalize_lang_code("rus"), "rus");
     }
 
     #[test]
@@ -1655,6 +1662,8 @@ mod tests {
             .collect();
         assert!(codes.contains(&"rus"));
         assert!(codes.contains(&"eng"));
+        assert!(!codes.contains(&"fra"));
+        assert!(!codes.contains(&"deu"));
     }
 }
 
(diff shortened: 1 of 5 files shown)


crates/remux-sdks/src/remux/mod.rs
```diff
@@ -5616,6 +5650,40 @@ impl StreamCodec {
...
 
+/// Common audio languages for the stream-group UI, alphabetical by display name.
+/// Key = ISO 639-2/B code (`MediaStream.language`), value = display name.
+/// `fre`/`fra` and `ger`/`deu` are both listed because ffprobe may emit either.
+pub fn common_audio_languages() -> &'static [(&'static str, &'static str)] {
+    &[
+        ("ara", "Arabic"),
+        ("chi", "Chinese"),
+        ("eng", "English"),
+        ("fre", "French"),
```
lostb1t, line 5696: i might be wrong but doesnt this render the same language twice in the dash?

crates/remux-server/src/db/stream_group.rs
```diff
@@ -402,6 +421,34 @@ impl StreamGroup {
                 let hit = values.contains(&codec);
                 matches!(op, SetOp::In | SetOp::Is) == hit
             }
+            // No probe data → no match. Unprobed sources must NOT be absorbed
+            // by the group (which only shows one representative); they fall
+            // through to the ungrouped list so they remain visible.
+            StreamRule::AudioLanguage { op, values } => match probe_data {
```
lostb1t: StreamRule::Size passes on None (unknown size) so HTTP/debrid sources aren’t silently dropped, AudioLanguage should do the same.

lostb1t (changes requested):

crates/remux-sdks/src/remux/mod.rs
OlegA-X: Yeah, you're right. Fixed

crates/remux-server/src/db/stream_group.rs
OlegA-X: Fair point.
Initially, with None => false because None => true was making unprobed sources disappear: filter_sources collapses all matches into a single representative (matching[0]) and adds the rest to matched_ids, so they never reached the ungrouped list. 
To make it consistent I had to refactor -matches() now returns a MatchOutcome enum (Match PassThrough/NoMatch). None => PassThrough (like Size). filter_sources only absorbs strict Matches into a group; PassThrough falls through to ungrouped so sources stay visible. Same for Resolution's existing Unknown case.

lostb1t:
ill merge when you fix the conflict

OlegA-X:
> ill merge when you fix the conflict

Done

PR #154, opened 2026-07-31
fix: use known runtime when ffprobe reports short duration

sean-wils:
## Scenario:
1. Open Streamyfin, try to play any MKV content sourced via MediaFusion/RealDebrid where the episode is longer than ~10 minutes
2. It'll fail with "no usable streams found"
3. Server logs show `stream is suspiciously short` with `probed_ticks=1209600000` (always exactly 120.96s)

The root cause is the fast probe flags (`-analyzeduration 0 -probesize 500000`).
ffprobe can't determine the real duration from MKV containers over HTTP with those limits, so it always reports ~121s. The short-duration check then rejects it because 121s < 5min threshold.

This PR is to fix that.

### DIFF
--- a/crates/remux-server/src/playback/engine.rs
+++ b/crates/remux-server/src/playback/engine.rs
@@ -675,14 +675,7 @@ pub(crate) fn build_hls_args(params: &TranscodeParams) -> Vec<String> {
             .map(VideoCodec::is_hevc)
             .unwrap_or(false);
 
-    let mut args: Vec<String> = vec![
-        "-v".into(),
-        "error".into(),
-        "-analyzeduration".into(),
-        "1000000".into(),
-        "-probesize".into(),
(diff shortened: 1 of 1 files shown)


lostb1t:
how about just using ffmpeg defaults? i think those are still fast enough 

the other issue with this is that upstreams often serve an custom short video when theres something wrong.

sean-wils:
Reworked. Just drops the restrictive `-analyzeduration/-probesize` flags and lets ffmpeg use its defaults. This fixes the false MKV duration reads without losing stub detection. Currently testing this.

sean-wils:
Sorry; this one has been sitting dormant for a while. I'll retest this now.

sean-wils:
Unable to recreate on the latest and greatest.

PR #155, opened 2026-08-01
feat: add healthcheck endpoint

squelix:
Just add a simple healthcheck endpoint for docker compose healthcheck.


README.md
```diff
@@ -70,6 +70,16 @@ services:
       /remux/data:/data
 ```
 
+A `/health` endpoint is available on port 3000 and returns `ok`, useful for a docker compose healthcheck:
```
lostb1t: removw from readme

crates/remux-server/src/api/system.rs
```diff
@@ -63,6 +63,12 @@ pub async fn system_ping(State(state): State<AppState>) -> Result<impl IntoRespo
     Ok(Json(json!("Remux Server")))
 }
 
+/// Healthcheck endpoint for orchestrators (e.g. docker compose)
+#[get("/health")]
+pub async fn health() -> &'static str {
+    "ok"
```
lostb1t: jellyfin says "Healthy" so we should too

lostb1t (changes requested):

PR #182, opened 2026-08-04
fix(music): scan and catalog local music with Jellyfin layout

### LINKED ISSUE #179: Local Music not found
Local Music saved at paths like

/media/Music/Steely Dan/Two Against Nature/03 - Two Against Nature.flac

/media/Music/Incubus - Morning View XXIII (2024) [FLAC] 88/03. Wish You Were Here.flac

Are not scanned in to the library when adding a local addon and can't be played by remux. Jellyfin sees these files and handles them no problem.

lostb1t:
fixes #179


lemonlambda:
any update?

Issue #186, opened 2026-08-05
Implement webhooks for Remux events


jonas-thuvesson-winberg:
Jellyfin has an official plugin for webhooks to be able to send notifications to external/third-party applications to track things like playback events.

It would be nice if Remux had the same capabilities (webhooks for events).

squelix:
i'm working on it ;)

PR #191, opened 2026-08-05
feat: webhooks

squelix:
# Try to implement the Webhooks #186 

remux doesn't have Jellyfin's plugin system and isn't going to get one, but the thing people actually use plugins for is webhooks piping library and playback events into Discord, n8n, Home Assistant, whatever. So this adds them natively.

The goal was to be a drop-in replacement for [jellyfin-plugin-webhook](https://github.com/jellyfin/jellyfin-plugin-webhook): same event names, same template variables, same Handlebars helpers. If you have a Discord template from the plugin, you should be able to paste it in and have it work.

## What's in it

Server events go through an internal broadcast channel to a background dispatcher, which filters per webhook, builds the payload dictionary, renders the operator's template and delivers over HTTP.

- **15 events**: ItemAdded, ItemDeleted, PlaybackStart/Progress/Stop, AuthenticationSuccess/Failure, SessionStart, TaskCompleted, UserCreated/Deleted/Updated/PasswordChanged/DataSaved, plus Generic for the test button. I skipped the plugin's six `Plugin*` events, PendingRestart and SubtitleDownloadFailure nothing in remux maps to them. UserLockedOut is out too, we don't have lockout.
- **Two destinations**: Generic (custom headers + fields) and Discord. The architecture takes a new one in an enum variant and a match arm, so Slack/Gotify/Pushover can follow later if anyone wants them. SMTP and MQTT would each drag in a real dependency, so they're not on my list.
- **Handlebars templates** per webhook, with the plugin's five helpers (`if_equals`, `if_exist`, `link_to`, `url_encode`, `json_encode`) and the `SendAllProperties` / `TrimWhitespace` / `SkipEmptyMessageBody` options.
- **Filtering** by event type, user and item type, same as the plugin.
- **CRUD API** under `/remux/webhooks` (additive, lowercase, doesn't touch the Jellyfin surface) plus a `/test` endpoint for the dashboard button.
- **Dashboard page** at Settings → Webhooks.

Delivery is fire-and-forget with a short retry: transport errors, 5xx, 408 and 429 get three attempts with exponential backoff, and 429 honours `Retry-After`. Fatal 4xx aren't retried. Deliveries are bounded to 4 in flight per webhook deliberately per-hook and not global, so one dead Discord URL can't starve your working Gotify hook.

## Deliberate deviations from the plugin

Worth knowing when comparing behaviour:

- **We escape for JSON, not HTML.** The C# plugin HTML-escapes and its stock templates use `{{{triple}}}` to opt out. Ours escapes `"`, `\` and control chars instead, so the shipped Discord template uses `{{double}}` otherwise a movie called `The "Burbs` produces invalid JSON and Discord 400s. `{{{triple}}}` is still there as the raw escape hatch.
- **We don't reproduce the plugin's colour bug.** Its `FormatColorCode` slices `hexCode[1..6]` and silently drops the last hex digit, so `#AA5CC3` renders as a different colour. Ours parses all six.
- **Generic picks a content type** instead of always sending `text/plain`. An explicit `Content-Type` header still wins.

Discord itself follows the plugin exactly: the destination options are injected as template variables and the operator's template renders the whole payload. No server-side envelope.

## New config

`public_url` the server's own externally reachable URL. The stock Discord template uses it for the thumbnail and the deep link, and it renders empty when unset (I'd rather ship an empty string than guess wrong). Note the env var is the bare `PUBLIC_URL`, which some environments already set.

## Drive-by fix

The test harness wasn't overriding `torrent_peer_port`, so every test server tried to bind in 6881..6891 and we were capped at ~10 concurrent. That's why `cargo test -p remux-server` needed `--test-threads=1`. Fixed the suite now runs in parallel, 10s instead of 79s.

## Screenshots 

Capture d’écran 2026-08-06 à 00 49 47
Capture d’écran 2026-08-06 à 00 49 52
Capture d’écran 2026-08-06 à 00 49 57


crates/remux-server/src/services/webhooks/template.rs
```diff
@@ -0,0 +1,729 @@
...
+/// The name the template is registered under while it is being checked. It
+/// appears in handlebars' error text, so it has to read as something the
+/// operator recognises rather than as an internal id.
+const VALIDATION_NAME: &str = "webhook template";
+
+/// Whether an operator-supplied template parses. The error text is derived from
+/// the operator's own template — never from a remote response — so it is safe
+/// to hand back over the API.
+pub(crate) fn validate(template: &str) -> Result<(), handlebars::TemplateError> {
+    Handlebars::new().register_template_string(VALIDATION_NAME, template)
```
sean-wils: `validate()` uses a bare `Handlebars::new()` while delivery uses `fresh_registry()`.

Helper typos pass save-time validation and silently drop every delivery at render time. Should use `fresh_registry()` in `validate()` too.

crates/remux-server/src/tasks/catalog_import_shared.rs
```diff
@@ -131,6 +134,20 @@ where
...
+        // are new. A scan can produce tens of thousands of them, so the whole
+        // loop is skipped — at the cost of one atomic load per chunk — when no
+        // webhook subscribes.
+        if ctx
+            .webhooks
+            .wants(NotificationType::ItemAdded)
+        {
+            for item in new_items.iter() {
+                ctx.webhooks
+                    .emit(WebhookEvent::ItemAdded { item_id: item.id });
```
sean-wils: Per-item emission can overflow the 4096-cap broadcast channel on large scans. `RecvError::Lagged` is only logged, so events are silently dropped. Worth batching `ItemAdded` instead.

crates/remux-server/src/services/webhooks/sender.rs
```diff
@@ -0,0 +1,1682 @@
...
+                dropped_since,
+                "webhook already has its share of deliveries in flight, dropping event"
+            );
+        }
+        return false;
+    };
+    tokio::spawn(async move {
+        // Held for the whole delivery, retries included: the slot is the
+        // ceiling on work owed to one endpoint, not on one HTTP round-trip.
+        let _permit = permit;
```
sean-wils, line 188: The permit is held across retries and `Retry-After` sleeps. In a 429 burst all 4 slots can be pinned sleeping, dropping new deliveries for the hook until the retries finish.

crates/remux-server/src/services/webhooks/mod.rs
```diff
@@ -0,0 +1,1042 @@
...
+                warn!(error = %e, "failed to load webhooks, keeping previous set");
+                self.inner
+                    .cache
+                    .write()
+                    .await
+                    .server = server;
+                // Ask for another attempt: the flag was consumed before the
+                // call, so nothing else will set it.
+                self.inner
+                    .dirty
```
sean-wils, line 399: `reload()` re-sets `dirty` and immediately retries on the next inbound event when DB access fails. During an outage, that effectively becomes one failing DB query per playback event. 💥 

Can we add retry backoff to reload failures?

crates/remux-server/src/services/webhooks/sender.rs
```diff
@@ -0,0 +1,1682 @@
...
+
+/// Delivery slots, counted **per hook**.
+///
+/// Not a single process-wide pool: one blackholing endpoint would hold every
+/// slot for its full retry window and drop deliveries to healthy hooks too. The
+/// total stays bounded at `enabled hooks × limit`.
+///
+/// TODO: entries are never removed. `WebhookService::reload` in `mod.rs` knows
+/// the live hook set and is the natural place to prune from.
+pub(crate) struct DeliverySlots {
```
sean-wils, line 96: `DeliverySlots` never gets pruned, so removed/rotated webhooks leave stale entries and the map can grow unbounded over time (memory leaks). Worth pruning on dirty reload (build active webhook keys, then `retain` matching entries).

crates/remux-dashboard/src/pages/webhooks.rs
```diff
@@ -0,0 +1,1605 @@
...
+/// Lower-case on purpose: `<input type="color">` round-trips its value in lower
+/// case and [`WebhookForm::to_dto`] stores the normalized form, so the swatch,
+/// the text field, the stored value and the placeholder are all one string.
+const DEFAULT_EMBED_COLOR: &str = "#3399ff";
+
+/// Every [`NotificationType`], in the order the SDK declares them.
+///
+/// Hand-written, so a variant added to the SDK must be added here too — the
+/// array's declared length is what pins the count.
+const NOTIFICATION_TYPES: [NotificationType; 15] = [
```
sean-wils: We can derive `EnumIter` on `NotificationType` and replace this const list with `NotificationType::iter()`. This keeps the checkbox list automatically in sync, so a new variant won't silently disappear from the UI.

crates/remux-server/src/services/webhooks/sender.rs
```diff
@@ -0,0 +1,1682 @@
...
+            body: rendered.to_string(),
+            content_type: HeaderValue::from_static(JSON_CONTENT_TYPE),
+            headers: Vec::new(),
+        },
+    }
+}
+
+/// The content type to send when the operator has not named one. A deviation
+/// from the plugin, which sends everything as `text/plain`.
+pub(crate) fn detect_content_type(body: &str) -> &'static str {
```
sean-wils, line 580: `detect_content_type` parses the full body as JSON just to set a header, then discards the result. Might be worth using a simple first non-whitespace byte check instead to avoid unnecessary parsing work.

crates/remux-server/src/services/webhooks/sender.rs
```diff
@@ -0,0 +1,1682 @@
...
+        return None;
+    }
+    Some(Duration::from_secs_f64(
+        seconds.min(MAX_RETRY_AFTER.as_secs_f64()),
+    ))
+}
+
+/// `base * 2^attempt` plus jitter in `[0, base/2)`, mirroring
+/// `remux_utils::retry!`.
+fn backoff(base_ms: u64, attempt: u32) -> Duration {
```
sean-wils: `backoff()` duplicates the `remux_utils::retry!` formula. Consider extracting to a shared utility to keep both in sync.

sean-wils (changes requested):

sean-wils:
Nice, looking good. I'll test it out.

squelix:
@lostb1t Do you want me to close this ? And we will do this later ?

lostb1t:
You can leave it open. Ill get to it eventually but currently other things have priority

PR #204, opened 2026-08-09
feat: Yamtrack scrobbling via general-purpose tracking addon system

sean-wils:
POC for foundational tracking addon with an initial implementation with Yamtrack

### DIFF
--- a/crates/remux-dashboard/src/pages/users.rs
+++ b/crates/remux-dashboard/src/pages/users.rs
@@ -2,10 +2,11 @@ use crate::{components::*, pages::streams::StreamFilterEditor, state::AppState};
 use dioxus::prelude::*;
 use remux_sdks::remux::{
     AddonDto, AdminSetPassword, CollectionFilter, CreateUser, DeleteUser, FilterGroup,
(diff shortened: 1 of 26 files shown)


OlegA-X:
It’s great that you’ve taken on the implementation. I was also thinking about scrobbling integrations, but I was wondering how to implement it so that users could manage their own integrations. I was thinking about a /profile (or /my) section - similar to the /admin section, but for a logged-in user. There, they could manage: their integrations (Yamtrack token, and in the future - Trakt, Simkl, Last.fm, etc.), and in the future, it could be expanded to include other personal settings.

What do you think about this? Doesn’t it go beyond the scope of the project’s idea and make things too complicated?@lostb1t

lostb1t:
While im not against the idea, it would be wise to discuss this first on discord to get the scope right

sean-wils:
Sounds wise. Good ideas @OlegA-X.
I'll drop a message in Discord.

sean-wils:
Closing this. Will cherry pick what I need when it comes to it.

sean-wils:
Superseded by https://github.com/lostb1t/remux/pull/304.

PR #227, opened 2026-08-14
fix: sweep orphaned rqbit torrents immediately on playback stop

### LINKED ISSUE #226: Torrent-sourced (Torznab/rqbit) downloads not cleaned up on stop or by Clear Cache
**Description**

When a stream is played from a torrent-backed source (Torznab indexer results, resolved via the built-in rqbit engine) rather than a debrid/HTTP addon, the downloaded file is never cleaned up — not when playback stops, and not via the "Clear Cache" action in settings.

**Reproduction**

1. Configure a Torznab-compatible indexer as a stream source (no debrid addon in the chain for that result).
2. Play a movie. Confirm in logs it resolves via rqbit (`cache/rqbit/<hash>.torrent` + `session.json` get created) and a real file appears under `data/torrents/<release name>/`.
3. Stop playback in the client.
4. Observe: the downloaded file remains at full size in `data/torrents/`, and the rqbit engine still shows open peer connections for the completed torrent — nothing was torn down when the play session ended.
5. Go to Settings → Clear Cache and run it.
6. Observe: the task log shows
   ```
   starting task=Clear Cache
   completed task=Clear Cache elapsed=52.4µs
   ```
   completing in ~50 microseconds — far too fast to have touched an 800MB+ media file. Confirmed the file in `data/torrents/` and the torrent state in `data/cache/rqbit/` are both untouched (same size/mtime before and after).

**Expected behavior**

Either:
- Playback stopping should tear down the torrent (or at minimum trigger cleanup after some idle/grace period), and/or
- "Clear Cache" should actually remove completed/orphaned torrent data from `data/torrents/` and the corresponding entries in `data/cache/rqbit/`.

**Actual behavior**

Downloaded torrent files accumulate indefinitely in `data/torrents/` with no automatic or manual (via the Clear Cache button) way to remove them from the UI. This is a real disk-usage and lingering-swarm-participation concern for anyone using Torznab/rqbit-backed sources instead of a debrid addon — the data note this comes from was a single ~810MB movie fully persisted with no cleanup path found.

**Environment**

- Image: `ghcr.io/lostb1t/remux:nightly`
- Stream source: Torznab indexer → rqbit (no debrid addon in that resolution path)
- Reproduced via the in-app Settings → Clear Cache action

Happy to provide more log excerpts if useful — redacted the ones above to just the relevant task-timing lines.

husnainahmedrana:
## Summary

Fixes #226.

Torznab/rqbit-resolved downloads (as opposed to debrid-backed sources) were only ever swept by the daily `CleanTranscodeFolder` scheduled task (`0 0 */24 * * *`), so a completed torrent-backed download sat at full size in `data/torrents/` — with rqbit still holding it in `data/cache/rqbit/` — for up to 24h after playback stopped. Neither stopping playback nor running "Clear Cache" touched it (Clear Cache is, by design, only the in-memory metadata/HTTP cache — not a bug on its own, just not what the disk usage needed).

This PR:
- Extracts the existing active-session/torrent-id bookkeeping + `delete_unused_with_files` call out of `CleanTranscodeFolderTask` into a shared `torrent::cleanup_unused(&ctx)` helper.
- Calls that helper (in the background, so it doesn't delay the HTTP response) right after `POST /sessions/playing/stopped` records a session stop, in `api::session::report_playback_stopped`.
- Keeps the daily `CleanTranscodeFolder` task running `cleanup_unused` too, as a safety net for sessions that never went through a clean stop (crash, killed client, etc).
- Updates the task's description so it's clear it also handles orphaned torrent downloads, not just transcode temp files.

## Test plan

- [ ] `cargo fmt` / `cargo build` (no Rust toolchain available in the environment I wrote this in — please run CI / a local build to confirm)
- [ ] Manual: play a Torznab/rqbit-resolved title, stop playback, confirm the release folder under `data/torrents/` and its entry in `data/cache/rqbit/` are removed within a few seconds instead of persisting until the next daily cron run
- [ ] Manual: confirm a torrent still being watched by another active session is *not* deleted when a different session referencing the same or a different torrent stops

### DIFF
--- a/crates/remux-server/migrations/202608140001_reduce_torrent_cleanup_interval.sql
+++ b/crates/remux-server/migrations/202608140001_reduce_torrent_cleanup_interval.sql
@@ -1,0 +1,11 @@
+-- Shrink the default CleanTranscodeFolder interval from 24h to 4h so
+-- orphaned Torznab/rqbit torrent downloads (see #226) don't sit around for
+-- up to a full day on installs that don't get the eager on-stop sweep a
+-- chance to run (e.g. a session that never reports a clean stop).
+-- Only touches rows still on the shipped default, so an operator's own
+-- customized interval is left untouched.
+UPDATE task_triggers
+SET cron = '0 0 */4 * * *'
+WHERE task_id = 'CleanTranscodeFolder'
+  AND kind = 'IntervalTrigger'
+  AND cron = '0 0 */24 * * *';
--- a/crates/remux-server/src/api/session.rs
+++ b/crates/remux-server/src/api/session.rs
@@ -24,6 +24,12 @@ use crate::{
     services::MediaResolveService,
 };
 
+/// How long to wait after a `playing/stopped` report before sweeping unused
+/// rqbit torrents, so a seek (which briefly detaches a session's transcode)
+/// or a quick rewatch of the same title isn't punished by an eviction that
+/// fired the instant the stop was reported.
+const TORRENT_CLEANUP_GRACE_PERIOD: Duration = Duration::from_secs(15 * 60);
+
 #[post("/sessions/logout")]
 pub async fn sessions_logout(
     State(state): State<AppState>,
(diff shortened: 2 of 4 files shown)


husnainahmedrana:
Pushed a follow-up commit reworking the approach after thinking through edge cases more:

- The on-stop sweep is now delayed 15 minutes (debounce) instead of firing instantly. Immediate deletion had two problems: (1) a seek briefly detaches a session's transcode while ffmpeg restarts at the new offset, and an immediate sweep triggered by a *different* session stopping at that exact moment could evict a torrent still being watched; (2) rewatching the same title a few minutes after stopping would force a full re-download even though the file was still there. `cleanup_unused()` re-snapshots active sessions at the moment it actually runs, so a 15-min delay means a resumed/rewatched session is seen as active again and skipped.
- Added a migration to shrink the `CleanTranscodeFolder` cron fallback from 24h to 4h, as a safety net for sessions that never report a clean stop (crash, killed client). It's a plain `UPDATE ... WHERE cron = '<old default>'`, so it won't clobber an operator's own customized interval.

Still unverified against a local `cargo build`/`cargo fmt` — no Rust toolchain available in the environment I wrote this in, flagged in the original PR description too.

husnainahmedrana:
Closing this without merging. Turns out the practical fix doesn't need code changes: `CleanTranscodeFolder`'s interval trigger is already user-configurable from Settings → Scheduled Tasks, so anyone hitting the disk-leak from #226 can just reschedule it from the default 24h down to something like 4h themselves, no upgrade required. Confirmed live that running the task does correctly clean up orphaned Torznab/rqbit torrents.

The on-stop debounced sweep in this branch would still shave that window down further (~15 min instead of waiting for the next cron tick), but it added real complexity to get the edge cases right (seek race, rewatch-soon-after), so it's not worth carrying for a marginal gain over "just lower the interval." Leaving the branch up at shanayela:fix/torrent-cleanup-on-stop in case the idea is useful later.

PR #237, opened 2026-08-16
fix: regenerate collection/library image when you remove it

### LINKED ISSUE #209: Images are not regenerated for collections
When removing an image from an existing collection it should regenerate it on save (when no other new image isnuploaded)

squelix:
Fixes #209

Removing a collection or library image doesn't give you the generated placeholder back, you just keep seeing the old picture. You mentioned it refreshes in the clients but not in the dashboard, that's true but it's not really a dashboard problem, the dashboard just happens to be where you notice it first. Three bugs stacked on top of each other.

**The image tag is a constant when there's no image row.** In `db_to_dto`, a collection or folder with no `media_images` row falls back to a synthetic `ImageTags.Primary` equal to `media.id`. That never changes. Images go out with `Cache-Control: max-age=86400` and no ETag, so same tag, same URL, and the browser happily replays the picture you just deleted for the next 24 hours. Timeline is: new collection → tag is `media.id` → browser caches the generated placeholder under that URL → you upload a poster (new row, new tag, fine) → you remove it → the row is gone so we're back to `media.id` → cache hit on the very first URL. It only looks like the dashboard because that's the client that shows you the edit form right after.

**Orphaned placeholder files.** `save_image` writes `{kind}.{ext}` following the upload format, but the generated placeholder is always `primary.jpg`. Upload a PNG over a generated placeholder and the JPEG stays behind, and `delete_image` only unlinked whatever path was in the DB row. So the orphan survives, and `library_image`'s `if path.exists()` check serves it forever. You get "a" generated image back, just the wrong one, possibly with the old library name baked in.

**`library_image` only recorded its row on one branch.** The `INSERT OR IGNORE` runs when it generates, not when it returns the existing file. So you can end up with a placeholder on disk and no row, which puts the DTO back on the synthetic tag permanently, and `process_image` keys its disk cache on the constant `placeholder:{id}` too.

## Changes

`delete_image` now sweeps every local `{kind}.{ext}` variant on top of the DB path, so nothing can shadow a later regeneration. `library_image` got restructured so both branches upsert, a row always exists when it returns. And a new `regenerate_library_image` deletes the current Primary then regenerates, which is what actually mints a new row UUID, and that UUID is the tag clients cache on. It's a no-op for anything that isn't a Collection or Folder.

It runs in three places: after a Primary delete in `delete_item_image_inner`, and at the two rename sites in `items.rs` that were already doing a bare `delete_image` for the same reason (the name is drawn into the image). Both of those already run after the title is committed, so the rebuilt placeholder picks up the new name.

Dashboard side, the Remove button now refetches the item to get the new tag and shows you the regenerated placeholder, instead of setting `has_image = false` and hiding the whole thing. `current_image_url` had to become a signal since it was computed once at page load and never recomputed.

Nothing changes on the public API, `DELETE /items/{id}/images/{type}` still returns 204 with no body, no new fields anywhere.

## The trade-off

Regeneration is synchronous, and `generate()` can go fetch a child backdrop over HTTP with a 15s timeout. The dashboard always sends `name` on save, so renaming a collection can now block for that long in the worst case with a dead backdrop URL. Before, the rename path was just a delete and the cost was pushed to the next image GET.

I did it inline on purpose, because that's the whole point, the fresh tag has to exist by the time the response comes back or the client has nothing new to request. Spawning it would make saves fast again but you'd be back to a stale tag for a moment. Happy to flip it if you'd rather have the latency, just say so.

Upload still works by the way, the dashboard does `PatchItem` then `UploadItemImage`, so the placeholder the rename regenerates gets immediately replaced by the actual upload under its own fresh UUID.

### DIFF
--- a/crates/remux-dashboard/src/pages/collections.rs
+++ b/crates/remux-dashboard/src/pages/collections.rs
@@ -448,13 +448,19 @@ pub fn CollectionForm(
         .server
         .manual_address
         .clone();
-    let current_image_url = existing_item_id
+    let existing_uuid = existing
+        .as_ref()
+        .map(|f| f.id);
+    let initial_image_url = existing_item_id
         .as_ref()
         .zip(existing_image_tag.as_ref())
         .map(|(id, tag)| format!("{server_base}/Items/{id}/Images/Primary?tag={tag}"));
+    let mut current_image_url: Signal<Option<String>> =
+        use_signal(|| initial_image_url.clone());
     let mut pending_image_bytes: Signal<Option<Vec<u8>>> = use_signal(|| None);
     let mut pending_image_preview: Signal<Option<String>> = use_signal(|| None);
     let mut has_image = use_signal(|| existing_image_tag.is_some());
+    let server_base_delete = server_base.clone();
     let client_for_delete = app_state.clone();
     let app_state_delete = app_state.clone();
     let delete_name = existing
@@ -711,7 +717,7 @@ pub fn CollectionForm(
                                 src: "{preview}",
                                 style: "width:100%;max-height:180px;object-fit:cover;border-radius:6px;border:1px solid var(--border)",
                             }
-                        } else if let Some(url) = &current_image_url {
+                        } else if let Some(url) = current_image_url.read().as_ref() {
                             if *has_image.read() {
                                 img {
                                     src: "{url}",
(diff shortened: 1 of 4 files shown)


lostb1t:
fixed in main

Issue #207, opened 2026-08-10
Tracking Integrations


lostb1t:
Implement support for external tracking services such as Trakt and Yamtrack.

### Connect tracking services

As a user, I want to connect and configure external tracking services, so that Remux can synchronize my watch activity with them.

Acceptance criteria:

- Users can connect supported tracking services such as Trakt and Yamtrack.
- OAuth-based services can be authenticated through their normal OAuth flow.

### Manage connected services

As a user, I want to see and manage my connected tracking services, so that I know which services are connected to my account.

Acceptance criteria:

- Users can see which tracking services are connected.
- Users can disconnect a tracking service.
- Users can reconfigure a connected service.
- Users can see whether a connected service is currently working.
- Sync failures expose enough information for the user to understand whether they need to fix their credentials or report a problem.

### Sync playback activity

As a user, I want my playback activity to be synchronized with all my connected tracking services, so that my watch history and progress stay up to date automatically.

Acceptance criteria:

- Playback progress is reported when playback is paused or stopped.
- Movies/episodes are marked as watched when playback is completed.
- Manually marking an item as watched also updates connected tracking services.
- Playback activity is sent to all connected tracking services automatically — no manual selection required.

### Choose which events to sync

As a user, I want to choose which events are sent to each connected service (e.g. only completed watches, or also progress updates and ratings), so that I have granular control over what each service receives.

Acceptance criteria:

- Users can configure per-service event filters (e.g. progress, completion, ratings, favorites).
- Changes take effect without reconnecting the service.

### Sync favorites and watchlist

As a user, I want favorites (and optionally watchlists) to sync both ways where the service supports it, so that items I save in one place appear in the other.

Acceptance criteria:

- Marking an item as a favorite in Remux pushes it to connected services that support favorites.
- Favorites added in a connected service are reflected in Remux.
- Watchlist sync follows the same two-way model where supported.

### Watch Later playlist

As a user, I want a personal "Watch Later" playlist created automatically when my account is created, so that I have a consistent place to save items across all integrations.

Acceptance criteria:

- A "Watch Later" playlist is created for every new user on account creation.
- Scrobbling integrations can sync their watchlist/queue to this playlist.
- The playlist behaves like any other personal playlist in Remux.

### Import tracking history

As a user, I want my existing watch history and progress to be imported when I connect a tracking service, so that I don't have to recreate my history manually.

Acceptance criteria:

- Connecting a service triggers an import of available watch history.
[...]

PR #238, opened 2026-08-16
fix: redact secrets from Debug output

sean-wils:
Taken from finding [lostb1t/remux#222 (comment)](https://github.com/lostb1t/remux/pull/222#discussion_r3791473499)

Fixing it properly.

Part of #207.


crates/remux-server/src/db/auth.rs
lostb1t, line 1: regression on line 487

let synthetic_device = Device {
    id: format!("apikey-{}", api_key.access_token),   // ← now prints "<redacted>"

lostb1t (changes requested):

sean-wils:
I think it's good to go now; I've retested and added some more coverage.

PR #239, opened 2026-08-16
fix: honour the selected stream group on Android TV

### LINKED ISSUE #220: Switching Stream Groups on the Jellyfin Android TV app issues
When Stream Groups are created,  the Android TV  app does not switch between them

The first group is always played

Switching to an Ungrouped stream will play after closing and reopening the App.

squelix:
Fixes #220

I went digging because "the first group is always played" sounded like a resolution bug on our side, but PlaybackInfo was actually fine all along selecting a group by its UUID already returns the right stream (there's even a test for it). The problem is one step earlier, in what we hand the client.

Here's what Android TV does when you pick a version in the details screen (`FullDetailsFragment` + `PlaybackController`, on master):

1. it refetches the item using **the MediaSource's Id**, and replaces its cached item with whatever comes back;
2. at play time it looks for the source whose `Id == item.Id`, and otherwise falls back to `mediaSources.get(0)`.

We fail both checks in the same way. `/Items/{group_uuid}` maps the group back to its owning item and returns the parent, with every version in priority order and `db_media_to_item` stamps `MediaSources[0].Id` with the item id. So whichever group you pick, the app lands on the highest-priority one. Switching versions could never work.

So: when the requested id is a stream group, that group's source is moved to the front and keeps its own UUID instead of being stamped with the item id. The client then sends the group back on PlaybackInfo and the existing (already tested) path takes over. All the other versions stay in the list so the version picker doesn't shrink after a switch.

Everything else is untouched the default listing still puts the item id on `MediaSources[0]`, which other clients rely on for auto-play.

### DIFF
--- a/crates/remux-server/src/api/items.rs
+++ b/crates/remux-server/src/api/items.rs
@@ -1556,8 +1556,13 @@ async fn item_for_user(
     let show_ungrouped = server_config
         .stream_groups_show_ungrouped
         .unwrap_or(true);
+    // Clients that switch versions (Android TV) refetch the item by MediaSource id
+    // and then play MediaSources[0], so the requested group must end up first and
+    // keep its own UUID instead of the item id stamped by `db_media_to_item`.
+    let mut requested_group: Option<Uuid> = None;
     let resolved_id = match MediaResolveService::resolve_item(id, &state.ctx).await? {
         Some(m) if m.kind == db::MediaKind::StreamGroup => {
+            requested_group = Some(m.id);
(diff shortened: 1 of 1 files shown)


lostb1t:
did you manual test this?

squelix:
> did you manual test this?

On emulator yes, but if someone  has a real Android TV it would be a better test 🙏

lostb1t:
aight nice work ill look at it more
close tomorrow

PR #240, opened 2026-08-16
feat(db): add user_media_trackers for per-user media tracking connections

sean-wils:
Part of #207.

### Manage connected services.

As a user, I want to see and manage my connected tracking services, so that I know which services are connected to my account.

#### Acceptance criteria:

- [ ] Users can see which tracking services are connected.
- [ ] Users can disconnect a tracking service.
- [ ] Users can reconfigure a connected service.
- [ ] Users can see whether a connected service is currently working.
- [ ] Sync failures expose enough information for the user to understand whether they need to fix their credentials or report a problem.

Storage only, so nothing is ticked yet. One row per (user, tracking addon) holding credentials, the event filter, and connection health. The API and integrations page tick them. Developer criteria are in #222.

Not `addon_users`: a row there restricts which addons the user can play from, and `set_user_addon_override` replaces the whole set, dropping the credentials with it.

### DIFF
--- a/crates/remux-server/migrations/202608160001_tracking_connections.sql
+++ b/crates/remux-server/migrations/202608160001_tracking_connections.sql
@@ -1,0 +1,27 @@
+-- A user's connection to one tracking addon.
+--
+-- Deliberately not addon_users: that table is the per-user addon *source
+-- override* list, so a row there restricts which addons the user can play
+-- from, and replacing the override set would drop the credentials with it.
+CREATE TABLE tracking_connections (
+    id              BLOB PRIMARY KEY NOT NULL,
+    addon_id        BLOB NOT NULL REFERENCES addons(id) ON DELETE CASCADE,
+    user_id         BLOB NOT NULL REFERENCES users(id)  ON DELETE CASCADE,
+    -- disconnected | connected | error | auth_expired
+    status          TEXT NOT NULL DEFAULT 'disconnected',
+    -- Opaque provider-defined JSON: a webhook token, or an OAuth triple.
+    credentials     TEXT NOT NULL DEFAULT '{}',
+    -- JSON array of TrackingEventKind. Empty means send nothing.
+    event_filters   TEXT NOT NULL DEFAULT '[]',
+    last_success_at DATETIME,
+    last_error_at   DATETIME,
+    last_error      TEXT,
+    -- retryable | permanent
+    last_error_kind TEXT,
+    created_at      DATETIME NOT NULL,
+    updated_at      DATETIME NOT NULL,
+    UNIQUE (addon_id, user_id)
+);
+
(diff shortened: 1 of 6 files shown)


crates/remux-server/migrations/202608160001_user_integrations.sql
```diff
@@ -0,0 +1,27 @@
+-- A user's connection to one tracking addon.
+--
+-- Deliberately not addon_users: that table is the per-user addon *source
+-- override* list, so a row there restricts which addons the user can play
+-- from, and replacing the override set would drop the credentials with it.
+CREATE TABLE user_integrations (
```
lostb1t: name too generic

crates/remux-server/src/db/user_integration.rs
```diff
@@ -0,0 +1,716 @@
...
+        if err.is_retryable() {
+            Self::Retryable
+        } else {
+            Self::Permanent
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, sqlx::FromRow)]
+pub struct UserIntegration {
```
lostb1t: also to generic. a integration to what?

lostb1t (changes requested):

crates/remux-server/src/db/user_integration.rs
sean-wils: Yeah valid; I've updated it so it's in reference to tracking service connections specifically.

crates/remux-server/migrations/202608160001_user_integrations.sql
sean-wils: Renamed to `tracking_connections`, so it lines up with `tracking_outbox` in #242.

crates/remux-server/src/db/user_integration.rs
sean-wils: Now `TrackingConnection`, in `db/tracking_connection.rs`, with `TrackingStatus` and `TrackingErrorKind` alongside it.

sean-wils:
Sorry, I amended rather than adding a commit, so the rename is not its own diff. The before/after is here: https://github.com/lostb1t/remux/compare/0935002...2d3d22a

lostb1t:
nitpicking but i still find the naming generic. track of what? connection to who?

how about user_media_tracker?

sean-wils:
> nitpicking but i still find the naming generic. track of what? connection to who?
> 
> how about user_media_tracker?

Haha yeah that's a lot better. It sounded a bit nefarious without context as well tbf.
Pushing a change now.

sean-wils:
Good call, `user_media_trackers` it is. Struct is `UserMediaTracker`. Left it as a separate commit so the rename is its own diff.

PR #242, opened 2026-08-16
feat(tracking): add the delivery queue and its retry worker

sean-wils:
Part of #207.

### Handle synchronization failures.

As a user, I want Remux to retry temporary synchronization failures, so that temporary service outages don't cause me to lose watch history.

#### Acceptance criteria:

- [x] Temporary failures are retried automatically.
- [ ] Permanent/authentication failures are surfaced to the user.
- [ ] Failed synchronization does not silently lose the playback event.

Retryable leaves the row pending, doubling from 30s to a 6h cap, swept every 5 minutes. A provider's `Retry-After` wins only if it asks for longer. Permanent stops immediately and marks the media tracker, since retrying can't fix bad credentials.

The table is a generic `delivery_queue` keyed by kind rather than a tracking-specific outbox, so webhooks can reuse the same retry machinery without a second worker.

Nothing enqueues yet: the handlers that produce these rows come with the next PR, along with the other two AC.


crates/remux-server/migrations/202608160002_media_tracker_outbox.sql
```diff
@@ -0,0 +1,38 @@
+-- Pending tracking deliveries, one row per (event, media tracker).
+--
+-- Written synchronously by the request handler before it returns, so a
+-- scrobble survives a crash, a restart, or the provider being down. An
+-- in-memory channel would drop these on lag.
+CREATE TABLE media_tracker_outbox (
```
lostb1t: we might wanna make this an generic queue so that we can use it for webhooks aswell in the future? 

something like

enum QueueKind {
   Tracker(TrackerPayload)
   Webhook(WebhookPayload)
}

struct DeliveryQueue {
id
kind: QueueKind
}

thoughts?

sean-wils: Yeah works for me; no harm in a shared queue.

crates/remux-server/src/db/delivery_queue.rs
```diff
@@ -0,0 +1,567 @@
...
+
+        let (status, next_attempt_at) = match err {
+            TrackingError::Permanent { .. } => {
+                (MediaTrackerOutboxStatus::FailedPermanent, now)
+            }
+            TrackingError::Retryable { retry_after, .. } => {
+                if attempts >= MAX_ATTEMPTS {
+                    (MediaTrackerOutboxStatus::FailedRetryable, now)
+                } else {
+                    let wait = backoff(attempts, *retry_after);
```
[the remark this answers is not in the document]
sean-wils: Fixed

PR #249, opened 2026-08-17
fix: preserve episode/season progress and watched state across a library purge

lostb1t:


squelix (approved):
Tested localhost and worked :)

PR #257, opened 2026-08-18
fix(jellyfin): align server routes and playback requests

veenix:
## Summary

- report `LocalAddress` from the request authority, including reverse-proxy headers
- serve the Jellyfin web client through both `/jellyfin` and the standard `/web` alias
- merge POST `PlaybackInfo` query parameters with the JSON body

These match request patterns used by Jellyfin web and WebOS clients and keep saved server identity/authentication associated with the endpoint the client actually reached.

## Testing

- `cargo fmt --all --check`
- `cargo check -p remux-server --lib`
- added coverage for the `/web` alias and request-derived `LocalAddress`

### DIFF
--- a/crates/remux-server/src/api/playback.rs
+++ b/crates/remux-server/src/api/playback.rs
@@ -54,9 +54,21 @@ pub async fn items_playbackinfo(
     State(state): State<AppState>,
     session: auth::AuthSession,
     Path(id): Path<Uuid>,
+    Query(query): Query<api::PlaybackInfoQuery>,
     Json(payload): Json<api::PlaybackInfoQuery>,
 ) -> Result<impl IntoResponse> {
-    items_playbackinfo_inner(state, session, id, payload).await
+    // Jellyfin web sends MediaSourceId as query param even for POST; merge query and body
+    let mut q = payload;
+    q.media_source_id = q
+        .media_source_id
+        .or(query.media_source_id);
+    q.user_id = q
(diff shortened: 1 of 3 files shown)


crates/remux-server/src/api/playback.rs
```diff
@@ -53,9 +53,27 @@ pub async fn items_playbackinfo(
     State(state): State<AppState>,
     session: auth::AuthSession,
     Path(id): Path<Uuid>,
+    Query(query): Query<api::PlaybackInfoQuery>,
     Json(payload): Json<api::PlaybackInfoQuery>,
 ) -> Result<impl IntoResponse> {
-    items_playbackinfo_inner(state, session, id, payload).await
+    // Jellyfin web sends MediaSourceId as query param even for POST; merge query and body
+    let mut q = payload;
+    if q.media_source_id
```
lostb1t: use shorthands

```

let mut q = payload;
q.media_source_id = q.media_source_id.or(query.media_source_id);
q.user_id = q.user_id.or(query.user_id);
q.device_profile = q.device_profile.or(query.device_profile);

```

lostb1t (changes requested):

crates/remux-server/src/api/playback.rs
veenix: Updated to use the `Option::or` shorthands, and rebased the PR onto the latest `main`.

PR #260, opened 2026-08-18
fix(api): playlist filtering, uuid format consistency and album runtime

MentallyOverwhelmed:
Heya! As a user of Remux, I really love this project, and one of the main reasons I started using it was due to the music features. because of this quite heavy use, i found some few bugs while using Feishin and Finamp as my clients and decided to fix them.

Here is what was happening:

* **fcacde5 & 9564bca:** Playlists were loading slowly, and duration data was missing (for example, NaNm NaNs in Feishin) or stuck at 0s (including album details). Remux was querying the database row by row and item by item. Now, parent references and runtimes are preloaded in batches (and album runtimes are calculated from child tracks), resulting in complete data and much faster load times.

* **22d2258:** When opening a playlist, non track items (like albums or artists) would sometimes show up. Feishin and Finamp explicitly ask for only tracks via `IncludeItemTypes`, but Remux was ignoring this filter and returning everything. Fixed by properly honoring `include_item_types`. This now fixes album/artist being shown as "Unknown artist" or "Unknown album"

* **7a474fd:** Users could accidentally add an entire artist or album to a playlist from Feishin, creating an unplayable, empty slot (which was shown in Finamp as a track, in clients like Feishin giving an "a is not iterable".) This helps Remux to silently reject anything that isn't a song track when adding playlist members. 
Note: This DOES NOT affect adding all the discography of an artist or all the discography of an album, if you add something like the "Michael Jackson" artist or the "Thriller" album, all the songs of the artist or album still get added.

* **5252c77:** Opening an album in Feishin (directly from the album search bar, opening albums directly from the track worked tho) would show "XX tracks" but the list showed up completely empty.  This was happening due to Remux returning the album ID without hyphens (`b9328cb1e906...`) but individual tracks with hyphens (`b9328cb1-e906...`). Feishin client does a strict string comparison and discarded them, this does not happen with Finamp. Unified everything into the standard hyphenated format.

* **769c749:** Browsing a genre in music clients would show the albums tab fine, but the songs tab was always completely empty, 0 songs. it was just excluding what the genre actually is. Remux just stores genres on albums (seems like the Deezer addon gets them from the album detail), this was making tracks to be excluded from genres. so when Feishin (and even finamp) asks for songs via the GenreIds filter nothing ever matched. Now tracks inherit the genres of their parent album, so the songs tab actually works. tested with indie pop, for example. now songs get shown and everything stays as Feishin and Finamp expect. 

Used Monochrome + Deezer addon for the tests.

### AI Disclosure
Following the repos policy, I want to disclose that the AI model used was Deepseek V4 Flash for this pull request via OpenCode to help map out the API and generate the initial fix. I have audited and read through the generated commits to understand the logic. Since the generated code was clean (and tested it in my own environment) I didn't need to manually change something in the code.

thanks for this really good fork of jellyfin! :D

*(Attached screenshots of Feishin and Finamp below as proof)*

### my own server (remux 0.25.0-nightly.20260818.g309405b)
{1830FB8E-F971-4051-B84B-B33D57D35CC9}
{8C3AE879-A268-4C42-9621-2D2F0226E028}
{64832931-6390-4F84-ADBB-9735E0752A2E}
{086E899D-8A86-4C47-85F9-6A1D672B0E33}

### Test server (built from source in Windows)
{F1BAAEAC-9C75-4771-86E9-6313756C42AE}
{6316E3C9-69D4-462B-A41F-29AEF63054CE}
{D0DC1135-95E0-439F-A3E5-49254B366526}
{7807B54D-B1D1-43FC-9FE0-97392DA5B0FF}


crates/remux-server/src/api/items.rs
```diff
@@ -94,6 +94,16 @@ impl ItemsQueryResultBuilder {
         self
     }
 
+    /// Fill `run_time_ticks` for Playlist rows in the payload before DTO
+    /// serialization. Playlists carry no runtime of their own in Remux;
+    /// Jellyfin clients expect the summed runtime of the member items.
+    pub async fn preload_playlist_runtimes(mut self, db: &sqlx::SqlitePool) -> Self {
```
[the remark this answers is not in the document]
lostb1t: adding to this. deezer should already have runtimes for albums?

lostb1t:
i think albums have runtimes in deezer no? so we can just save it on insert.

MentallyOverwhelmed:
> i think albums have runtimes in deezer no? so we can just save it on insert.

yesss you were right!! deezer gives us runtime info 

by the way, I wonder if Copilot is kind of outdated on the 999 limit? iirc it's 32766 in newer vers, we're using 3.46.0 so this is kind of odd. i pushed that commit anyways. in any case, if copilot is correct about that 999 for the variable limit, I added it and checked it,. Things added are:

1. create a helper called Media::get_by_ids in db/media.rs that chunks IN (...) in batches and merges results. It was applied to playlists.rs in retain_playlist_tracks (the flow of adding all songs from an artist/album with >999 tracks if there's a reason if someone has a lot of songs or if someone has a REALLLY big playlist to import via API or Deezer playlist ID), get_playlist_items (for the /Playlists/{id}/Items endpoints), items.rs when browsing playlists and preload_playlist_runtimes in db/media.rs so we can chunk both playlists and albums IN's. 

2. deezer albums do come with a total duration directly from the API. adding duration to the sdks dz::Album/dz::ArtistAlbumRef and persisted it as a runtime in the three insert paths that are fetch_album_meta, list_artist_albums and sync_artist_children. I kept the preloads for existing rows as a fallback, if you want me to remove them there is no problem. 

3. docstring is corrected, it now mentions albums too.

{05E0EAAB-63DF-4CA7-8392-A6C1B8B14902}

here it shows 41m 40s with the Deezer runtime.
changes are now pushed

crates/remux-server/src/api/playlists.rs
```diff
@@ -64,6 +97,7 @@ pub async fn create_playlist(
     if !ids.is_empty() {
         let resolved =
             crate::services::MediaResolveService::resolve_ids(&ids, &state.ctx).await;
+        let resolved = retain_playlist_tracks(&state.ctx.db, resolved).await?;
```
lostb1t: would this not prevent creation of playlists for video?

MentallyOverwhelmed: ooh oh yup, I just thought about music playlists, not about the other item ids.

sooo doing a POST /Playlists with a Movie gets created, but shows a "TotalRecordCount: 0", it's completely empty. same with POST /Playlists/{id}/Items with a Movie, nothing gets added, while doing this same POST with a Track gets added and shows a TotalRecordCount: 1. (that repro was on this commit release, before this fix). 
Movie, Episode, Track and TvChannel  MediaKind's can get added, Series, Season, Album, Artist, Collection, Folder, Playlist, StreamGroup  MediaKind's can not now.

for container items like an Album, Artist or Shows, i'd say it's the clients job to crumble them into episodes or tracks: the API doesn't do it, neither on prod nor on this commits. if a client sends the container id, the playlist gets created but the container gets filtered out, so it ends up empty

sooo that's why i renamed the filter to retain_playlist_items, with a new is_playable_leaf() helper on MediaKind. it should now allow all playable leaf items like Movies, Episodes, Tracks, TV Channels, while keeping our container items like Albums, Artists, TV Shows out of the playlists and leaving that to /collections.

feishin:
{7D56B255-B573-4682-8178-4E737E5BC417}
this one is a test playlist with a Track and a Movie now. Feishin will show as "2 music tracks" and the full runtime (of both the movie and Track) tho, but that's expected as Feishin does the item and runtime count.

fladder:
{888DC0F6-60BF-43FC-AE0A-B5CDC11C02D8}
it shows the track example and movie example correctly now.

the only thing that I can think about now is the UX experience: currently if a client sends only a container id like an album, the filter drops it, returning a 200/204 as if everything was ok, with an empty playlist and just a warn! log on the server. it's not critical tho. 
and i guess the runtime preload / item handling might be off for mixed kind like movies and tracks playlists too, but anyways, if it was me i'd rather have a separate playlist for my own music and my movies/episodes.

and damn you were right xd, it's tough to have good user friendly experience with music in Jellyfin.

PR #269, opened 2026-08-19
fix(addons): import Stremio catalogs whose type is the wildcard `all`

### LINKED ISSUE #267: Catalogs with Stremio type `all` are silently skipped by Refresh Library
### Summary

Catalogs whose Stremio `type` is not one of the mapped kinds are dropped from `Refresh Library` without a log line or an error. The catalog can be enabled in the dashboard, it stays enabled, and it simply never imports.

The concrete case is `type: "all"`, which AIOMetadata uses for every Trakt list catalog. On my install that is 14 of 199 catalogs, including lists like `Latest releases` and `Latest Dolby Vision Releases`. Any collection built on one of them renders as a permanently empty row.

### Version

`0.25.0-nightly.20260819.g524e084`

### Reproduce

1. Add an addon that publishes a catalog with `"type": "all"`. AIOMetadata does this for Trakt lists.
2. Enable that catalog in the dashboard.
3. Run `Refresh Library`.
4. No items import. No warning appears in the logs.

The addon itself is fine:

```
GET <aiometadata>/catalog/all/trakt.list.5308818.json  ->  50 metas
GET <aiometadata>/catalog/all/trakt.list.23552998.json ->  50 metas
```

Every meta in both responses carries its own `type` field (`movie` in these two samples), so the per item kind is available even though the catalog level type is `all`.

### Root cause

All line numbers are against `524e084`.

1. `crates/remux-server/src/db/media.rs:211-233`. `TryFrom<sdks::stremio::MediaType> for MediaKind` maps `Unknown(s)` for `episode`, `season` and `person` only. Everything else falls to `_ => Err(())` at line 229. `all` lands there.

2. `crates/remux-server/src/addons/stremio.rs:286`. The result is discarded with `.ok()`, so the catalog is registered with `media_kind: None`.

```rust
media_kind: db::MediaKind::try_from(stremio_kind).ok(),
```

3. `crates/remux-server/src/addons/mod.rs:1655-1662`. `catalogs_for_kinds` drops any catalog whose `media_kind` is `None`.

```rust
.filter(|c| {
    c.media_kind
        .as_ref()
        .map_or(false, |k| kinds.contains(k))
})
```

4. `crates/remux-server/src/tasks/refresh_library.rs:55-66`. `LIBRARY_KINDS` has no variant that an `all` catalog could ever match, so the filter above removes it before any request is made.

The catalog is never attempted, which is why nothing is logged.

`type: "collection"` hits the same path, and so would any custom type such as `anime`. `custom_stremio_type` at `db/media.rs:239` already preserves the raw type string for exactly that reason, so the concept is present, it just does not reach the refresh filter.

### Suggested fix

Two options, in order of preference.

1. Treat `all` as a fan out. Import the catalog and derive each item's kind from the meta's own `type` field rather than from the catalog's. The data is already in the response.

2. Failing that, log a warning when an enabled catalog is skipped for an unmappable type, and mark it in the dashboard. Today the only symptom is an empty row, which is very hard to trace back to a type mismatch.

Happy to test a patch against my install.

squelix:
Fixes https://github.com/lostb1t/remux/issues/267

Catalogs whose Stremio `type` is `all` were dropped by `Refresh Library` without a log line. They stay enabled in the dashboard and just never import, so any collection built on one renders as a permanently empty row with nothing pointing back to the cause.

This is the `#267` case: AIOMetadata publishes every Trakt list as `type: "all"`.

## Why it happened

`MediaKind::try_from(MediaType::Unknown("all"))` returns `Err`, `.ok()` collapsed that into `media_kind: None`, and the filter in `catalogs_for_kinds` used `map_or(false, ...)` so the catalog was removed before any request was made, silently.

The item path was never the problem: every meta in an `all` catalog response carries its own `type`, and `TryFrom<stremio::Meta> for Media` already uses it. So the fix lives entirely at the catalog-admission layer.

## What changed

- `Option<MediaKind>` on `CatalogInfo`/`ResolvedCatalog` becomes a `CatalogKind` enum, so a provider can say *wildcard* (`ItemDefined`) as opposed to *nothing declared* (`Unspecified`). No new `MediaKind` variant that enum is the DB text encoding.
- `catalogs_for_kinds` takes an explicit `WildcardCatalogs::{Include,Exclude}`. `RefreshLibrary` claims wildcard catalogs; `RefreshIptv` asks for `TvChannel` alone and must not double-import the movies and series such a catalog also yields.
- A wildcard catalog's own channels are skipped at import. `prune_stale_iptv_channels` deletes any `tv_channel` older than its run regardless of source, so such a row would otherwise flap in and out of the library on every pair of refreshes. This puts the previously unused `_catalog` param of `import_catalog_items` to work.
- Other unmappable types (`collection`, `anime`, …) deliberately stay out: their items carry that same unmapped type and `TryFrom<stremio::Meta>` falls back to `Movie`, so fanning them out would import junk rows as films. They now log a warning when the catalog is enabled which is what was missing all along. That's option 2 from the issue, for the cases option 1 can't safely cover.

## Compatibility

`media_kind` is never persisted (`CatalogState` only stores enabled/max_items/tags) and `provider_catalog_id` keeps its `all:{id}` format, so catalogs users already enabled stay enabled. No migration.

### DIFF
--- a/crates/remux-server/src/addons/deezer.rs
+++ b/crates/remux-server/src/addons/deezer.rs
@@ -11,8 +11,8 @@ use uuid::Uuid;
 
 use super::{
     AddonCapabilities, AddonKind, AddonMetadata, AddonOption, AddonOptionType,
-    AddonPreset, AddonPresetRegistration, CatalogAddon, CatalogInfo, MediaKind,
-    MetaAddon, ResourceType, SearchAddon, TreeAddon,
+    AddonPreset, AddonPresetRegistration, CatalogAddon, CatalogInfo, CatalogKind,
+    MediaKind, MetaAddon, ResourceType, SearchAddon, TreeAddon,
 };
 use crate::{
     AppContext, common, db,
@@ -1047,7 +1047,7 @@ impl CatalogAddon for DeezerAddon {
                     _ => default_playlist_name(&id),
                 };
                 CatalogInfo {
-                    media_kind: Some(db::MediaKind::Playlist),
+                    media_kind: CatalogKind::Single(db::MediaKind::Playlist),
                     ..CatalogInfo::new(id, name)
                 }
             }
--- a/crates/remux-server/src/addons/iptv.rs
+++ b/crates/remux-server/src/addons/iptv.rs
@@ -8,8 +8,8 @@ use uuid::Uuid;
 
 use super::{
     AddonCapabilities, AddonKind, AddonMetadata, AddonOption, AddonOptionType,
-    AddonPreset, AddonPresetRegistration, CatalogAddon, CatalogInfo, MediaKind,
-    ProgressReporter, ResourceType, StreamAddon,
+    AddonPreset, AddonPresetRegistration, CatalogAddon, CatalogInfo, CatalogKind,
+    MediaKind, ProgressReporter, ResourceType, StreamAddon,
 };
 use crate::{AppContext, db, iptv};
 
@@ -335,7 +335,7 @@ impl CatalogAddon for IptvAddon {
             default_enabled: true,
             default_max_items: Some(999999999),
             collection_media_kind: None,
-            media_kind: Some(db::MediaKind::TvChannel),
+            media_kind: CatalogKind::Single(db::MediaKind::TvChannel),
         }])
     }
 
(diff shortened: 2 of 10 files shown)


lostb1t:
@squelix i appreciate you want to help. But i have to ask you to stop creating vibe coded PRs like this.

PR #270, opened 2026-08-19
fix(jellyfin-import): fix auth, paginate user items, and resolve episode states positionally

lostb1t:


kats1123:
There is still an unresolved episode/season UUID collision in this PR's fallback path.

`MediaIdRaw::canonical()` returns `None` for `Season` and `Episode`, so this line still maps every item that misses positional lookup to the same stable UUID:

```rust
positional.unwrap_or_else(|| uuid::Uuid::from(&raw))
```

I reproduced the underlying problem on current `main` using a Jellyfin-compatible adapter backed by a full Trakt export:

- 649 watch events
- 582 unique items
- the task logged `states_imported=582` and `states_unresolved=0`
- the database gained only 26 distinct rows: 25 resolved items plus one fallback row repeatedly overwritten by the unresolved items

The positional lookup in this PR will resolve more existing library items, but any episode absent from the local media table, or any positional lookup miss, still takes the colliding fallback.

I fixed and verified it locally by deriving the fallback UUID from `SeriesProviderIds` plus the season/episode position, matching Remux's flat Stremio IDs:

- episode: `stable_media_uuid(Episode, "{series_key}:{season}:{episode}")`
- season: `stable_media_uuid(Season, "{series_key}:{season}")`
- `series_key` preference: IMDb, then `tmdb:{id}`, then `tvdb:{id}`

After restoring the pre-import database and rerunning with that fallback:

- 582/582 distinct states matched
- 0 missing
- 0 play-count mismatches
- 0 timestamp mismatches
- all 67 repeat watches were preserved

A regression test using `tt35636246`, season 1, episode 1 produced the existing Remux UUID `8a3af6af-9e0f-51a4-9df3-99a8479ad04a`.

PR #271, opened 2026-08-19
fix(items): restore episode backdrop fallback

veenix:
## Summary

- treat `BackdropImageTags` as the non-optional vector defined by the API model
- fall back to optional parent backdrop tags without changing the intended Infuse behavior
- restore formatting and compilation after #262

## Testing

- `cargo test -p remux-server --lib -- --test-threads=1` (440 passed)

### DIFF
--- a/crates/remux-server/src/api/models.rs
+++ b/crates/remux-server/src/api/models.rs
@@ -965,14 +965,17 @@ pub fn db_media_to_item(media: db::Media, hide_sources: bool) -> BaseItemDto {
             db::ImageKind::Backdrop,
         )
         .map(|b| vec![b]);
-        
+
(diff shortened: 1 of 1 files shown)


veenix:
#262 introduced code that treated:
  - backdrop_image_tags: Vec<String> as an Option
  - parent_backdrop_image_tags: Option<Vec<String>> as a plain Vec

That caused three Rust type errors plus a formatting failure, breaking main and every subsequent PR’s CI.

This diff uses:
```
if item.backdrop_image_tags.is_empty() {
    item.backdrop_image_tags =
        item.parent_backdrop_image_tags.clone().unwrap_or_default();
}
```

So it restores compilation and formatting. All 440 server tests then pass.

PR #261, opened 2026-08-18
feat(tracking): queue watch activity as it happens

sean-wils:
Part of #207.

### Sync playback activity.

As a user, I want my playback activity to be synchronized with all my connected tracking services, so that my watch history and progress stay up to date automatically.

#### Acceptance criteria:

- [x] Playback progress is reported when playback is paused or stopped.
- [x] Movies/episodes are marked as watched when playback is completed.
- [x] Manually marking an item as watched also updates connected tracking services.
- [x] Playback activity is sent to all connected tracking services automatically.

Nothing is delivered until a provider exists. That is the next PR.

PR #272, opened 2026-08-19
refactor(tracking): build the delivery target at delivery

sean-wils:
Part of #207, follow-up to #242.

The target was built when an event was queued. It needs to be built when the event is delivered, because filling in an episode's missing ids means a network call, and that can't happen in a request handler.

Second commit fixes a bug where one undeliverable item marked the whole tracker broken.

### DIFF
--- a/crates/remux-server/src/addons/tracking.rs
+++ b/crates/remux-server/src/addons/tracking.rs
@@ -188,34 +188,39 @@ pub struct TrackingTarget {
     pub kind: db::MediaKind,
     pub title: String,
     pub year: Option<i32>,
-    pub ids: TrackingIds,
+    pub ids: db::ExternalIds,
     /// Set for episodes: the parent series' title, year and ids.
     pub series: Option<Box<TrackingTarget>>,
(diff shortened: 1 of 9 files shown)


sean-wils:
Sorry about this one @lostb1t. Pain in the arse discoveries as I began the initial implementation of Yamtrack to test things out.

crates/remux-server/src/addons/tracking.rs
```diff
@@ -600,29 +605,75 @@ mod tests {
...
+            db::ExternalIds {
+                tvdb: Some(79126),
+                ..Default::default()
+            },
+        )));
+        assert!(episode.is_matchable());
+    }
+
+    /// A music id identifies nothing to a tracking service, so it must not
+    /// stand in for one.
```
sean-wils, line 665: Maybe a TODO for the future; music tracking is out of scope for now.

sean-wils:
Not completely happy with it yet.

sean-wils:
Folded into #261; status fix split out as #279.

PR #273, opened 2026-08-19
fix(jellyfin): preserve session message protocol shapes

veenix:
## Summary

- restore Jellyfin enum casing for remote playstate commands after route normalization
- include the expected `LibraryUpdateInfo` body in `LibraryChanged` websocket messages
- add focused serialization and command-mapping tests

## Testing

Validated on top of the compile repair in #271:

- `cargo test -p remux-server --lib jellyfin -- --test-threads=1` (5 passed)

This PR remains independent; current `main` CI is blocked by the regression repaired in #271.

### DIFF
--- a/crates/remux-server/src/api/session.rs
+++ b/crates/remux-server/src/api/session.rs
@@ -831,6 +831,22 @@ struct RemotePlaystateQuery {
     controlling_user_id: Option<String>,
 }
 
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, strum_macros::Display, strum_macros::EnumString,
+)]
+#[strum(serialize_all = "PascalCase", ascii_case_insensitive)]
+enum PlaystateCommand {
+    Stop,
+    Pause,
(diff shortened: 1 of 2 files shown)


crates/remux-server/src/api/session.rs
```diff
@@ -831,6 +831,25 @@ struct RemotePlaystateQuery {
     controlling_user_id: Option<String>,
 }
 
+fn canonical_playstate_command(command: &str) -> String {
```
lostb1t: use enum and strum for this

lostb1t (changes requested):

crates/remux-server/src/api/session.rs
veenix: Updated to a typed `PlaystateCommand` using `strum` `EnumString`/`Display`, with case-insensitive parsing and canonical PascalCase output. Invalid command values now return 400. The focused command-casing and websocket payload tests pass.

PR #277, opened 2026-08-19
fix(torrent): stream only the selected bundle file

veenix:
## Summary

- initialize torrents with an empty file selection so metadata lookup does not download or allocate every file in a bundle
- resolve the requested video from an exact filename, a validated file index, or an unambiguous main feature
- reject ambiguous bundles instead of silently streaming file zero
- clear persisted file selections and bound finite HTTP ranges only when librqbit returns a partial response
- start paused torrents without suppressing unrelated start failures

## Why

Bundle torrents can contain several similarly sized movies, samples, metadata, and subtitles. The previous path could trust an invalid index, fall back to the first file, or leave files selected for background download. Finite client ranges were also forwarded to librqbit even though its stream endpoint accepts only open-ended ranges.

## Tests

- `cargo test -p remux-server torrent::tests`
- `cargo test -p remux-server torrent_proxy`
- `cargo clippy -p remux-server --all-targets -- -D warnings`
- `cargo test -p remux-server -- --test-threads=1` (453 library tests and 1 binary test passed; 1 doc test ignored)

## AI Disclosure

### DIFF
--- a/crates/remux-server/src/stream.rs
+++ b/crates/remux-server/src/stream.rs
@@ -2,11 +2,11 @@ use crate::ResultExt;
 use async_trait::async_trait;
 use axum::{body::Body, http::HeaderMap, response::Response};
 use axum_anyhow::ApiResult as Result;
-use futures_util::TryStreamExt;
+use futures_util::{StreamExt, TryStreamExt};
 use nutype::nutype;
 use std::{io, path::PathBuf};
 use tokio::io::{AsyncReadExt, AsyncSeekExt};
-use tokio_util::io::ReaderStream;
+use tokio_util::io::{ReaderStream, StreamReader};
 use uuid::Uuid;
 
 use crate::AppState;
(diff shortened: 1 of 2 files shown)


crates/remux-server/src/torrent.rs
```diff
@@ -111,31 +107,51 @@ impl TorrentManager {
...
+            None,
+        );
+        api.api_torrent_action_update_only_files(
+            TorrentIdOrHash::Id(torrent_id),
+            &std::collections::HashSet::new(),
+        )
+        .await
+        .context("failed to clear torrent file selection")?;
+        if let Err(error) = api
+            .api_torrent_action_start(TorrentIdOrHash::Id(torrent_id))
```
lostb1t: Any error will fall through. maybe theres a way to check state before hand or match on error kind if available.

crates/remux-server/src/stream.rs
```diff
@@ -336,18 +355,54 @@ impl StreamSource for HttpSource {
...
+            .and_then(|(_, total)| {
+                total
+                    .parse::<u64>()
+                    .ok()
+            });
+        let bounded_range = requested_range
+            .filter(|(start, end)| {
+                end.is_some()
+                    && (upstream_status == http::StatusCode::PARTIAL_CONTENT
+                        || *start == 0)
```
lostb1t: not sure why the start condition, dont we only wanna bound when theres an partial response, not matter the start range?

crates/remux-server/src/torrent.rs
veenix: Updated. The code now checks `handle.stats().state` and only calls start when the torrent is not already live. If another request starts it during the race, that specific case is suppressed; every other start error is returned with context.

crates/remux-server/src/stream.rs
veenix: Agreed. I removed the `start == 0` exception, so finite responses are bounded only when the upstream actually returned `206 Partial Content`. I also added a regression test confirming an upstream `200 OK` stays a full response.

PR #278, opened 2026-08-19
fix(subtitles): expose matching torrent sidecars

veenix:
## Summary

- discover matching `.srt` sidecars from already initialized torrent metadata through the stream abstraction
- associate filename-matched subtitles with bundle entries and allow generic subtitle directories only for unambiguous single-video releases
- expose sidecars through PlaybackInfo and download only the subtitle file selected by the client
- keep device- and media-source-scoped subtitle routes so injected indexes resolve consistently
- preserve language, forced, and hearing-impaired metadata from common subtitle filenames
- keep add-on subtitle indexes aligned when torrent sidecars are inserted first

## Why

Torrent releases often carry subtitles as separate files rather than embedded tracks. Those tracks were invisible to Jellyfin clients, and selecting one must not start every file in a bundle. The matching rules are deliberately conservative so a movie or episode does not receive a subtitle belonging to another bundled video.

## Tests

- `cargo test -p remux-server --lib -- --test-threads=1` (486 passed)
- `cargo clippy -p remux-server --lib -- -D warnings`

## AI Disclosure

### DIFF
--- a/crates/remux-server/src/api/playback.rs
+++ b/crates/remux-server/src/api/playback.rs
@@ -1,7 +1,10 @@
 use anyhow::anyhow;
 use axum::Json;
 
-use super::subtitles::{inject_external_subtitles, scored_external_subtitles};
+use super::subtitles::{
+    inject_external_subtitles, inject_sidecar_subtitles, save_sidecar_subtitle_routes,
+    scored_external_subtitles,
+};
 use axum::{
     body::Body,
     extract::{Path, State},
@@ -264,6 +267,11 @@ async fn items_playbackinfo_inner(
             .results
             .len(),
     );
+    let mut sidecar_subtitle_routes = Vec::with_capacity(
+        probed
+            .results
+            .len(),
+    );
 
     // Per-user playback preferences + remembered selections, resolved per source
     // via `MediaSourceInfo::resolve_default_streams` (see below).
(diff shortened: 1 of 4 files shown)


crates/remux-server/src/api/playback.rs
```diff
@@ -453,6 +461,19 @@ async fn items_playbackinfo_inner(
             TranscodeDecision::Transcode(outcome) => outcome.apply_to(&mut source),
         }
 
+        let sidecars = effective_stream
```
lostb1t, line 478: move sidecars getter to the stream struct

crates/remux-server/src/api/subtitles.rs
```diff
@@ -257,6 +351,43 @@ async fn subtitles_stream_inner(
     stream_index: i64,
     format: String,
 ) -> Result<impl IntoResponse> {
+    let manifest_key = sidecar_subtitle_manifest_key(
```
lostb1t: move this manifest handling to seperate free function

crates/remux-server/src/api/subtitles.rs
```diff
@@ -249,6 +249,100 @@ async fn fetch_external_subtitle_bytes(
...
+        .status(StatusCode::OK)
+        .header("Content-Type", content_type)
+        .header("Cache-Control", "public, max-age=3600")
+        .header("Access-Control-Allow-Origin", "*")
+        .body(Body::from(converted))
+        .unwrap()
+}
+
+#[derive(Clone)]
+pub(crate) struct SidecarSubtitleManifestEntry {
```
lostb1t: i know the word manifest is pretty generic but its something central to stremio addons. I suggest using something different

lostb1t (changes requested):

veenix:
Addressed the review feedback and rebased onto current main:

- sidecar discovery now lives on `StreamInfo`
- cached sidecar routing is handled by dedicated free functions outside the subtitle request handler
- renamed the sidecar `manifest` terminology to `routes` to avoid confusion with Stremio manifests
- squashed the PR to one commit

Full server library suite passes serially (486/486), and strict library Clippy passes.

PR #295, opened 2026-08-21
fix(ci): restore server compilation and formatting

veenix:
## Summary

- return track album IDs as `Uuid` values to match `BaseItemDto.album_id`
- apply the current rustfmt output to recently merged playlist and media code

## Why

Current `main` fails the server build because the album ID conversion still produces `Option<String>` after the API field changed to `Option<Uuid>`. Its formatting check also fails on the recently merged playlist/media changes. These two focused commits restore both required checks without changing runtime behavior beyond the corrected album ID type.

## Tests

- `cargo check -p remux-server --lib`

## AI Disclosure


lostb1t:
hmm i need to fix the ci then. this should not have been allowed to pass the checks. tnx!

PR #298, opened 2026-08-21
feat(tracking): complete an episode's external ids on demand

sean-wils:
Towards #207.

An episode is stored with a TMDB id alone, so a tracker matching on imdb or tvdb can't identify one. `complete_episode_ids` fills those in at delivery, off the playback path, and writes them back. It fills the series' tmdb id too, for trackers that key an episode on the show instead.


crates/remux-server/src/addons/mod.rs
```diff
@@ -1796,6 +1811,74 @@ impl AddonService {
         }))
     }
 
+    /// Ask every applicable meta addon to complete `media`'s external ids, and
+    /// persist the result. Returns whether anything was added.
+    #[tracing::instrument(skip_all, fields(title = %media.title, kind = %media.kind))]
+    pub async fn complete_external_ids(
```
lostb1t: we already have some functions on MediaResolveService that resolve ids

next to that im pretty sure we have an even better one somwhere that also falls back to kirsu and others because anime might not have an tmddb / id

we should probaoly leverage that. i dont think this belongs on the addons

lostb1t (changes requested):

crates/remux-server/src/addons/mod.rs
sean-wils: Sweet! Let me take a look.

sean-wils: Moved both lookups onto `MediaResolveService`, next to `resolve_media_imdb`. `addons/` is untouched now.

sean-wils:
Looks like my latest change broke something. I'll look at this later.

sean-wils:
#356 took this in along with the #308 refactor, so there's nothing left here.

PR #302, opened 2026-08-21
perf(streams): refresh stale cached sources asynchronously

veenix:
## Summary

- return existing cached torrent sources without waiting for provider refreshes
- refresh stale torrent-only source sets in a background task
- keep cold loads synchronous when no usable cached source exists
- refresh HTTP/debrid and mixed source sets synchronously so expiring URLs are not served stale
- serialize refreshes per media item and re-check freshness after acquiring the lock

## Why

After the stream cache TTL expires, item requests currently wait for every configured provider even when the server already has durable cached torrent descriptors. That adds provider latency to detail-page loading and playback controls. Serving those descriptors while revalidating removes the blocking delay without changing cold-item behavior.

The optimization is intentionally limited to torrent-only caches. HTTP and debrid URLs may expire, so those sources continue to refresh before they are returned.

## Tests

- focused torrent-only versus HTTP/debrid cache policy test
- full remux-server suite passes on the integration stack: 581 passed
- `cargo fmt --all --check`
- `cargo clippy -p remux-server --all-targets -- -D warnings`

## AI Disclosure

### DIFF
--- a/crates/remux-server/src/addons/mod.rs
+++ b/crates/remux-server/src/addons/mod.rs
@@ -2753,47 +2753,29 @@ fn match_probe_version<'a>(
         })
 }
 
+const STREAMS_TTL_SECS: i64 = 60;
+
+fn streams_are_fresh(
+    refreshed_at: Option<chrono::NaiveDateTime>,
+    now: chrono::NaiveDateTime,
+) -> bool {
+    refreshed_at.is_some_and(|refreshed_at| {
+        (now - refreshed_at).num_seconds() < STREAMS_TTL_SECS
+    })
+}
+
+fn can_refresh_cached_streams_in_background(total: i64, p2p: i64) -> bool {
+    total > 0 && total == p2p
+}
+
 impl AddonService {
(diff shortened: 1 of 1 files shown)


crates/remux-server/src/addons/mod.rs
```diff
@@ -2963,6 +2945,87 @@ impl AddonService {
...
+        .bind(media.id)
+        .fetch_one(&ctx.db)
+        .await
+        .unwrap_or((0, 0));
+        if can_refresh_cached_streams_in_background(cached_counts.0, cached_counts.1) {
+            let service = self.clone();
+            let mut media = media.clone();
+            let ctx = ctx.clone();
+            tokio::spawn(async move {
+                let _guard = STREAM_LOCKS
```
lostb1t, line 2980: think we can keep this dry; the refresh logic seems to be the same for background and live. Probably worth moving it into the inner function.

lostb1t (changes requested):

PR #303, opened 2026-08-21
fix(streams): prefer advertised seeder counts

veenix:
## Summary

- deserialize the optional Stremio stream title
- recover explicitly advertised seeder counts from stream text when structured metadata is absent
- use the highest advertised seeder count when choosing one representative for a stream group
- retain provider order as the tie-breaker

## Why

Some Stremio addons expose torrent health only in the stream name, description, or title. Remux currently discards the title and can choose the first matching grouped source even when another candidate advertises substantially more seeders.

Structured seeder metadata remains authoritative. The fallback parser only accepts the person marker or explicit seeders/seeds labels, and it performs no network probes or downloads.

## Tests

- full remux-server library suite: 497 passed
- cargo fmt --all -- --check
- cargo clippy -p remux-server --lib -- -D warnings

## AI Disclosure

### DIFF
--- a/crates/remux-sdks/src/stremio/mod.rs
+++ b/crates/remux-sdks/src/stremio/mod.rs
@@ -837,6 +837,7 @@ pub struct Stream {
     pub parsed_file: Option<ParsedFile>,
     pub name: Option<String>,
     pub description: Option<String>,
+    pub title: Option<String>,
     pub behavior_hints: Option<BehaviorHints>,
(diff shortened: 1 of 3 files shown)


lostb1t:
the thing with this is, i kinda expect users to use aiostreams. aio has a ton of features for filtering snd sorting so remux is hands off.

So we really cant just change order. if we do we need to gate it behind a setting. but that need some thought.

PR #304, opened 2026-08-21
feat(tracking): add webhook media trackers, with Yamtrack

sean-wils:
Towards #207.

Several trackers are fed through Jellyfin's webhook plugin and differ only in the body they send. `webhook_media_tracker` is the shared half: build the notification, post it, classify the reply. A provider supplies a `WebhookFormat` and nothing else.

Yamtrack is the first on it. Scrob follows in its own PR once this lands.

FYI connecting a tracker has no API yet, that'll be a separate PR, so this does nothing for users until that lands.

### DIFF
--- a/crates/remux-server/src/addons/media_tracker.rs
+++ b/crates/remux-server/src/addons/media_tracker.rs
@@ -10,6 +10,11 @@ use super::AddonKind;
 use async_trait::async_trait;
 
 /// Split by what the dispatcher should do next, not by cause.
+///
+/// The message is stored as `last_error` on the connection and returned by
+/// `GET /users/{id}/mediatrackers`, so it must never carry a credential. That
+/// is easy to do by accident: a provider that authenticates by url puts the
+/// secret in every transport error it raises.
 #[derive(Debug)]
 pub enum MediaTrackerError {
     /// Rate limited, 5xx, network. `retry_after` is a provider hint; the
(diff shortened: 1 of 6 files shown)


sean-wils:
Tested against a real Yamtrack instance:
1. episodes and movies both scrobble
2. an abandoned film records as in progress
3. an episode TMDB can't fill in fails visibly instead of reporting success

I'll look into fixing the unresolved episodes in a separate PR.

lostb1t:
i don't think webhooks and trackers are the same thing tho. wanted to tackle webhooks acfer this

do these trackers dont have an api?

sean-wils:
They do, but they don't follow a standard and would rely on us maintaining every self hosted tracking service under the sun in isolation if they change anything. They all share the same jellyfin webhook plugin functionality so in my mind it makes way more sense to me to just follow the standard defined where at all possible.

sean-wils:
If you''d rather tackle webhooks properly maybe let's close this for now and revisit once it's ready.

lostb1t:
i didnt gave it much thought but seeing your pr you provide the webhooks/body which makes sense

tho yeah i need to have a proper look again because webhooks in its core are generic and used outside of trackers

sean-wils:
Yeah, agreed they should be generic. I think the split is:
- a webhook is operator-level: any URL, any body
- a tracker connection is per-user, holds a credential, surfaces a reauth prompt when it goes stale, has a per-provider event filter, and the body is fixed so the user can't get it wrong

Yamtrack wants the second one, so I'd keep it on the tracker trait even once generic webhooks exist.

Renamed it to `media_tracker/jellyfin_webhook_body.rs` so it reads as the body and not as a claim on the area. Have a look whenever.

I'll keep this in draft until you'd have a think about it. 👍

lostb1t:
posted an idea on discord

PR #305, opened 2026-08-22
feat(streams): improve torrent source labels

veenix:
## Summary

- replace repetitive torrent source names with compact release details
- include source type, codec, HDR, audio, release group, and advertised seeders when available
- keep synthetic Auto labels unchanged
- leave local, HTTP, debrid, RTSP, and OpenDAL source names unchanged

## Why

Multiple torrent results at the same resolution are otherwise difficult to distinguish in Jellyfin clients. The additional metadata makes manual source selection useful without changing playback behavior.

## Tests

- full remux-server suite: 497 passed
- cargo fmt --all --check
- cargo clippy -p remux-server --all-targets -- -D warnings

Screenshot 2026-08-21 at 9 48 39 PM

## AI Disclosure

### DIFF
--- a/crates/remux-server/src/conversions.rs
+++ b/crates/remux-server/src/conversions.rs
@@ -85,6 +85,222 @@ fn infer_audio_channels(text: &str) -> Option<i64> {
     }
 }
 
+pub(crate) fn media_source_display_name(source: &db::Media) -> String {
+    if is_auto_source_name(&source.title) {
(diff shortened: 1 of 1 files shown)


lostb1t:
yeah this is my bad for not mentioning. aiostreams does all of this already. So we have to think twice if we want to replicate this functionality.

PR #313, opened 2026-08-22
feat: add native BetterPosters addon

telenice:
### **Summary**

Adds BetterPosters as a native Remux addon and makes it available as the effective poster provider for movies and series.

The integration is designed to work across normal library browsing and search results rather than only exposing BetterPosters through the manual image picker.

### **What changed**
Added a native BetterPosters addon under the existing Remux addon system.
Added configuration for BetterPosters poster options, including:
trend tags
quality tags
genre
rating
rating source
age rating
watch progress
language
BetterPosters posters are resolved at image request time instead of replacing the original stored artwork.
Original artwork remains available as a fallback if BetterPosters cannot return an image.
Remote search results are supported as well as media already stored in the library.
Reuses Remux's existing TMDB/IMDb resolution where an IMDb ID is not already available.
Added optional authenticated image requests so poster generation can use the requesting Jellyfin user's local playback state.
Local watch state is used for watched/progress poster variants instead of requiring Trakt or PublicMetaDB.
Primary image tags include BetterPosters configuration/watch state so Jellyfin clients request updated artwork when poster settings or playback state change.
User-specific posters use private caching to avoid sharing watch-state artwork between users.

### **Implementation notes**

The BetterPosters addon does not overwrite the media's primary image in the database.

Instead, the image endpoint resolves an effective poster when a client requests Primary or Thumb artwork:

Resolve the media from the database or Remux's cached search results.
Resolve an IMDb ID, using the existing TMDB mapping when needed.
Build the BetterPosters URL from the addon configuration.
Include local user watch state when watch-progress overlays are enabled.
Proxy the resulting image through Remux.
Fall back to the existing Remux artwork path if BetterPosters is unavailable.

Keeping this at image-request time means disabling the addon immediately restores the original artwork and avoids permanently modifying library metadata.

The image cache tag is also derived from the relevant BetterPosters settings and local watch state. This is needed because Jellyfin clients cache posters by ImageTag; without changing the tag, toggling an overlay or changing watched state can leave an older generated poster visible.

### **Testing**

Tested with:

cargo fmt --all
cargo check -p remux-server

I also built and ran the modified server in Docker and verified BetterPosters artwork is returned through the Jellyfin UI while changing addon configuration.

### **AI disclosure**

AI was used to help research the Remux addon/image paths and draft parts of the initial implementation. I reviewed and edited the resulting code, compiled and tested it against my Remux checkout, and worked through the integration and runtime issues before submitting this PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -451,7 +451,7 @@ jobs:
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
-          username: lostb1t
+          username: telenice
           password: ${{ secrets.GH_TOKEN }}
 
       - name: Build and push by digest
@@ -498,7 +498,7 @@ jobs:
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
-          username: lostb1t
+          username: telenice
           password: ${{ secrets.GH_TOKEN }}
 
       - name: Set up Docker Buildx
--- a/crates/remux-server/src/addons/better_posters.rs
+++ b/crates/remux-server/src/addons/better_posters.rs
@@ -1,0 +1,546 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use std::{
+    collections::hash_map::DefaultHasher,
+    hash::{Hash, Hasher},
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
+};
+use uuid::Uuid;
+
+use super::{
+    AddonCapabilities, AddonKind, AddonMetadata, AddonOption, AddonOptionType,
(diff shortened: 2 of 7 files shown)


telenice:
AIOMetadata maybe, not AIOStreams though. AIOMetadata is quite resource heavy and I don't use it for my catalogs. That's why I think this is a simple addon to replace posters in library for existing catalogs and search

 

 

  <https://avatars.githubusercontent.com/u/168401?s=20&v=4> lostb1t left a comment (lostb1t/remux#313) <https://github.com/lostb1t/remux/pull/313#issuecomment-5384732394> 

dont aiostreams / aiometadata support this

PR #318, opened 2026-08-23
fix(search): accept camelCase searchTerm in hints

Jasshl:
## Summary

- accept camelCase `searchTerm` on `/Search/Hints`
- preserve support for canonical `SearchTerm`
- add regression coverage for both spellings

## Reproduction

Litefin sends `/Search/Hints?searchTerm=Naruto`. Remux 0.26.0 treated the parameter as absent and returned no title results, while the same request using `SearchTerm` returned movies and series.

## Validation

- `cargo test -p remux-sdks` (50 passed)
- direct lowercase `/Search/Hints` request returned movie and series results
- manually verified search in Litefin against the patched build.

### DIFF
--- a/crates/remux-sdks/src/remux/mod.rs
+++ b/crates/remux-sdks/src/remux/mod.rs
@@ -4346,6 +4346,7 @@ pub struct RemoteImageResult {
 #[derive(Debug, Default, Deserialize)]
 #[serde(rename_all = "PascalCase")]
 pub struct SearchHintsQuery {
+    #[serde(alias = "searchTerm")]
     pub search_term: Option<String>,
     pub start_index: Option<u32>,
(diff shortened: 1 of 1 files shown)


crates/remux-sdks/src/remux/mod.rs
```diff
@@ -4350,6 +4346,7 @@ pub struct RemoteImageResult {
 #[derive(Debug, Default, Deserialize)]
 #[serde(rename_all = "PascalCase")]
 pub struct SearchHintsQuery {
+    #[serde(alias = "searchTerm")]
```
lostb1t, line 4349: theres a query struct level macro you can use for this

lostb1t (changes requested):

PR #326, opened 2026-08-24
fix(music): resolve search clicks to the clicked item, add artist art…

OlegA-X:
Clicking a track or album from music search opened the artist page instead of the item - the resolve path persisted and returned the artist root, and for tracks on compilations (soundtracks, VA releases) that never appear in the artist's discography this was permanent. Playback of freshly searched tracks hit the artist's streams for the same reason. This PR makes the click resolve to the clicked item itself (syncing its album when the release is missing), moves the heavy artist discography sync to the background so first clicks are faster.
Аnd adds Artist support to the Deezer meta addon so artists now get their name and photo on persist.


lostb1t:
this needs a rebase

PR #359, opened 2026-08-26
feat: drag and drop polish

jonas-thuvesson-winberg:
- feat(dashboard): auto-scroll page while dragging list items
  - Near the edge of the top or bottom of the page. Especially helpful on long lists of collections.
- fix(dashboard): show loading state after reordering addons and collections
  - Previously (if the update was slow) you could continue to re-order items and then suddenly the page refresh would cut you short
  
Disclosure: Codex helped me create the edge scroll solution on drag.

### DIFF
--- a/crates/remux-dashboard/src/components/drag_and_drop_list/component.rs
+++ b/crates/remux-dashboard/src/components/drag_and_drop_list/component.rs
@@ -5,6 +5,9 @@ use dioxus_primitives::drag_and_drop_list::{
     DragAndDropListItemProps, DragAndDropListItemsProps,
 };
 
+const AUTO_SCROLL_EDGE_PX: f64 = 96.0;
+const AUTO_SCROLL_MAX_STEP_PX: f64 = 12.0;
+
 #[css_module("/src/components/drag_and_drop_list/style.css")]
 struct Styles;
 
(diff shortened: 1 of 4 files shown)


lostb1t:
why does it need a page refresh?

jonas-thuvesson-winberg:
You're right it doesn't... just kept the pattern from before when i did the initial implementation. It only needs to update via the api. I still think it should show the loading state though, since it could cause issues if you have a slow connection, and try multiple changes in short succession.

I can remove the refresh trigger.

jonas-thuvesson-winberg:
I'll try to make the list non-interactive during the request instead, to avoid the flicker/rerender

jonas-thuvesson-winberg:
Added an interaction state for the drag-and-drop list to lock it during updates, and removed the full page refresh entirely.

Codex with manual edits used.

jonas-thuvesson-winberg:
Removed the interaction locking after some discussion with @lostb1t regarding how it impedes the user experience. We will leave that part, and revisit if it becomes a problem. A more user friendly implementation could be a background queue (if that is needed in the future).

jonas-thuvesson-winberg:
Added error handling and fixed a bug where the reordering would update all items instead of only the items reordered.

Codex used for the solutions, and reviewed and tested by me.

PR #360, opened 2026-08-27
feat(stremio): keep the tvdb id an episode arrives with

sean-wils:
Towards #207.

Cinemeta sends `tvdb_id` on every video and it was dropped twice: the field was never deserialised, and `stremio_meta_episode` rebuilt `external_ids` with only the stremio id on it.

That is the only id an episode row carries that names it outside remux. Over 150,000 episodes in the library I tested had none.

Tested: The Bob Hope Show went from 0 of 268 episodes with a tvdb id to 56, matching what Cinemeta has.

Only helps if Cinemeta is configured. AIOStreams and AIOMetadata both send no tvdb ids.

#363 is the follow-up for episodes no addon carries an id for.

### DIFF
--- a/crates/remux-sdks/src/stremio/mod.rs
+++ b/crates/remux-sdks/src/stremio/mod.rs
@@ -798,6 +798,13 @@ impl Meta {
 #[serde(rename_all = "camelCase")]
 pub struct Episode {
     pub id: String,
+    /// The episode's own TVDB id, which Cinemeta carries per video and TMDB's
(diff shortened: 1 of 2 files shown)


sean-wils:
This isn't a perfect solution. It resolved about 50% of the episodes to scrobble to Yamtrack.
Will also test whether syncing TheTVDB will help.

lostb1t:
Superseded by #442, a rebase onto current main that resolves the merge conflicts.

PR #365, opened 2026-08-27
fix(api): improve Jellyfin API and DTO compatibility for Roku

voidxela:
This PR improves compatibility with the Jellyfin Roku client across our API and DTO models. It updates `ProfileCondition.value` deserialization to accept numbers, booleans, and strings, resolving payload parsing errors when the Roku client sends numeric condition properties like `AudioChannels: 6`. It also removes `skip_serializing_if` from `BaseItemDto` list fields so empty collections are serialized as empty JSON arrays rather than omitted, matching Jellyfin's expected API contract and preventing client-side property lookup failures.

*Co-authored and verified with Antigravity AI assistant.*

### DIFF
--- a/crates/remux-server/src/api/models.rs
+++ b/crates/remux-server/src/api/models.rs
@@ -1069,14 +1069,6 @@ pub fn db_media_to_item(media: db::Media, hide_sources: bool) -> BaseItemDto {
             .as_ref()
             .and_then(|s| s.first())
             .and_then(|s| s.bitrate);
-        item.media_streams = item
-            .media_sources
-            .as_ref()
-            .and_then(|s| s.first())
-            .map(|s| {
-                s.media_streams
-                    .clone()
-            });
         if media.kind != db::MediaKind::Track {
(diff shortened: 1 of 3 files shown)


crates/remux-server/src/api/models.rs
```diff
@@ -1069,6 +1069,14 @@ pub fn db_media_to_item(media: db::Media, hide_sources: bool) -> BaseItemDto {
             .as_ref()
             .and_then(|s| s.first())
             .and_then(|s| s.bitrate);
+        item.media_streams = item
```
lostb1t: this will get out of sync because media sources gets mutated down the pipeline. Remove if not needed

lostb1t (changes requested):

crates/remux-server/src/api/models.rs
voidxela: Removed the  assignment. Clients read streams directly from `MediaSources[i].MediaStreams`, so this was unnecessary.

PR #368, opened 2026-08-27
fix: accept trailing slash for items search route

voidxela:
This PR allows the `/items` API route to accept requests with a trailing slash (`/items/`). The Jellyfin Roku client issues search and item listing requests to `/Items/` with trailing slashes, which previously failed routing. Registering both `/items` and `/items/` ensures full compatibility with the Roku client without impacting existing client requests.

*Co-authored and verified with Antigravity AI assistant.*

### DIFF
--- a/crates/remux-server/src/api/items.rs
+++ b/crates/remux-server/src/api/items.rs
@@ -867,7 +867,7 @@ pub async fn items_flat(
     Ok(Json::<Vec<api::BaseItemDto>>(items.items))
 }
 
-#[get("/items", "/items/")]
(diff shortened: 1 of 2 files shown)


lostb1t:
trailing slash should be handled globally. Add it to rewrite_request_uri

voidxela:
Done, handled trailing slash globally and added unit tests with edge cases. Needed exceptions for mounted SPA roots (/web/, /admin/, /jellyfin/) which expect a trailing slash and would create an infinite redirect loop.

PR #374, opened 2026-08-28
fix(sdks): redact client error request context

skynet01:
## Summary

SDK request errors can include their endpoint and response body in log output. For example, an add-on endpoint with a token in its query string can be exposed when a failed request is formatted by tracing.

This keeps the status and safe error message while redacting endpoint and body fields in both normal and debug formatting. Transport and URL-construction errors now report their category without formatting opaque request context.

## Testing

- `cargo test -p remux-sdks error_formatting_redacts_credentialed_endpoints`

### DIFF
--- a/crates/remux-sdks/src/lib.rs
+++ b/crates/remux-sdks/src/lib.rs
@@ -96,34 +96,94 @@ impl Auth for JellyfinApiKeyAuth {
     }
 }
 
-#[derive(Debug, thiserror::Error)]
+#[derive(thiserror::Error)]
 pub enum ClientError {
     #[error("unauthorized")]
     Unauthorized,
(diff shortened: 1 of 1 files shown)


crates/remux-sdks/src/lib.rs
lostb1t, line 1: We have an existing Secret<T> wrapper in remux-utils for exactly this purpose (redacted Debug, greppable .expose() reads). Rather than a manual fmt::Debug impl, the cleaner fix is to change the two fields to Option<Secret<String>>

lostb1t (changes requested):

crates/remux-sdks/src/lib.rs
skynet01: No problem, sorry about that didnt know that existed

PR #379, opened 2026-08-28
fix(transcode): SW tonemap when QSV VPP + subtitle burn-in; preserve tonemap on SW fallback

### LINKED ISSUE #378: HW-accelerated (QSV) transcode fails when subtitles are burned into HDR10 content. Falls back to software encoding without tone mapping
When transcoding an HDR10 (HEVC Main10) file with QSV hardware acceleration enabled and subtitles being burned into the video (overlay), ffmpeg fails with a filter-graph error and Remux falls back to software encoding. The software fallback does not apply tone mapping, so HDR content plays back visibly washed out (colours flat, subtitle text rendering grey instead of white) instead of erroring out or applying tone mapping in software too.

With subtitles disabled, hardware QSV transcoding (including hardware VPP tone mapping) works correctly and looks right.

**Environment**

- Remux version: 0.27.0 (ghcr.io/lostb1t/remux:latest, image digest sha256:f2de64aa9f5601c4bbe94ae2bdf15d49bd99bbb2aed0503d60ac29f33d56622a)
- Host: Docker Compose on Ubuntu, kernel 7.0.0-28-generic
- GPU: Intel Rocket Lake (Gen12) iGPU, /dev/dri/card1
- VA-API driver: iHD (/usr/lib/jellyfin-ffmpeg/lib/dri/iHD_drv_video.so), VA-API version 1.23.0
- Settings: Hardware VPP tone mapping (VAAPI/QSV) enabled under Playback settings
- Source file: MKV container, HEVC Main 10, HDR10, TrueHD 7.1, 117.8 GiB, ~93.6 Mbps
- Client: bundled Jellyfin Web player in a browser (forces full transcode — container, video codec, audio codec, HDR range, and subtitle codec are all reported as unsupported by this client)

**Steps to reproduce**
- Enable Hardware VPP tone mapping (VAAPI/QSV) in Remux's playback settings and restart the container.
- Play an HDR10 HEVC file with an embedded subtitle track selected/forced, from a client that requires subtitle burn-in (e.g. the bundled web player).
- Observe the transcode falls back to software encoding, and the picture is washed out / not tone-mapped.
- Play the same file with subtitles set to None.
- Observe hardware QSV transcoding with tone mapping now succeeds and looks correct.

**Expected behavior**

- Hardware-accelerated transcoding with tone mapping succeeds regardless of whether subtitles are being burned in
OR
- If the hardware filter chain genuinely can't support burning subtitles into a tone-mapped HW pipeline, the software fallback should still apply the same tone-mapping conversion (in software) rather than silently skipping it, so HDR content doesn't render washed out.

**Actual behavior**

The hardware transcode attempt fails with a filter-graph error, and the software fallback plays back HDR content without any tone mapping applied. Colours are washed out and on-screen text (subtitles) renders grey instead of white, consistent with HDR metadata being reinterpreted as SDR without an actual tone curve conversion.

This happened identically on three separate playback attempts.

**Evidence**

ffmpeg failure (from docker logs remux)
```
WARN remux_server::playback::engine:1126: HW-accelerated transcode failed — retrying with software encoding accel=Qsv stderr="libva info: VA-API version 1.23.0
libva info: Trying to open /usr/lib/jellyfin-ffmpeg/lib/dri/iHD_drv_video.so
libva info: Found init function __vaDriverInit_1_23
libva info: va_openDriver() returns 0
libva info: VA-API version 1.23.0
libva info: Trying to open /usr/lib/jellyfin-ffmpeg/lib/dri/iHD_drv_video.so
libva info: Found init function __vaDriverInit_1_23
libva info: va_openDriver() returns 0
Impossible to convert between the formats supported by the filter 'Parsed_overlay_3' and the filter 'auto_scale_0'
[fc#0 @ 0x...] Error reinitializing filters!
[fc#0 @ 0x...] Task finished with error code: -38 (Function not implemented)
[fc#0 @ 0x...] Terminating thread with return code -38 (Function not implemented)
[vost#0:0/h264_qsv @ 0x...] Could not open encoder before EOF
[vost#0:0/h264_qsv @ 0x...] Task finished with error code: -22 (Invalid argument)
[vost#0:0/h264_qsv @ 0x...] Terminating thread with return code -22 (Invalid argument)
[out#0/hls @ 0x...] Nothing was written into output file, because at least one of its streams received no packets."
```

This repeated on three separate playback attempts (~09:03–09:04), each time falling back to software encoding after the identical filter error.

intel_gpu_top comparison shows the GPU/driver are otherwise fine

With subtitles ON (playback "working" via the software fallback):

Image

GPU at 100% RC6 (fully idle) — confirms nothing is running on the GPU; the "working" playback is pure software.

With subtitles OFF (same file, same settings, subtitle track set to None):

Image

GPU clocked to 1278/1295 MHz, ffmpeg process visible and active (PID 2362644). VideoEnhance engine usage confirms the hardware VPP tone-mapping step is genuinely running, and the result looks correct.

This isolates the bug to the specific combination of hardware-accelerated (QSV/VAAPI) transcoding plus subtitle burn-in (not the hardware, driver, or tone-mapping feature individually, both of which work correctly on their own).

Happy to provide the full uncut logs, the exact source file's mediainfo, or test further combinations if useful.

--

This report was drafted with Claude's help based on my own logs and testing (including isolating the issue via the intel_gpu_top comparison above). I reviewed it before submitting. Hopefully this falls in line with the AI policy. Thanks!

lostb1t:
fixes #378

### DIFF
--- a/crates/remux-server/src/playback/engine.rs
+++ b/crates/remux-server/src/playback/engine.rs
@@ -705,10 +705,19 @@ pub(crate) fn build_hls_args(params: &TranscodeParams) -> Vec<String> {
                      [vmain][sub]{overlay_qsv}[v]"
                 )
             } else {
+                // When do_vpp_tonemap is active but we're in the CPU overlay path,
+                // tonemap_vaapi (hw_suffix) expects VAAPI frames, but overlay outputs
+                // CPU yuv420p — incompatible. Switch to software tonemapping and use
+                // the CPU-compatible hw suffix (format=nv12) for the encoder instead.
+                let (effective_sw_tonemap, effective_hw_suffix) = if do_vpp_tonemap {
+                    (true, accel.hw_filter_suffix(hdr, false))
+                } else {
+                    (do_sw_tonemap, hw_suffix.clone())
+                };
                 let main_video = build_cpu_main_video_filters(
                     build_scale_filter(params),
                     hdr,
-                    do_sw_tonemap,
+                    effective_sw_tonemap,
                     &params.tonemapping_algorithm,
                     params.tonemapping_desat,
                     params.tonemapping_peak,
@@ -717,7 +726,7 @@ pub(crate) fn build_hls_args(params: &TranscodeParams) -> Vec<String> {
                     sub_idx,
                     &sub_preproc,
                     &main_video,
-                    hw_suffix.as_deref(),
+                    effective_hw_suffix.as_deref(),
                 )
             };
             args.extend(["-filter_complex".into(), filter]);
@@ -1161,6 +1170,11 @@ pub async fn start_transcode(
                 let _ = std::fs::remove_dir_all(&params.output_dir);
                 let _ = std::fs::create_dir_all(&params.output_dir);
                 params.accelerator = Box::new(NoAccel);
+                // VPP tonemap can't run without HW; preserve the user's intent by
+                // activating software tonemapping so HDR content isn't washed out.
+                if params.enable_vpp_tonemapping {
+                    params.enable_tonemapping = true;
+                }
                 sw_fallback = true;
                 continue;
             }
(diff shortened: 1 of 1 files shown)


lostb1t:
Superseded by #408 — the filter-string fix here left the decoder still requesting VAAPI hw frames while the filter graph switched to CPU-only filters, producing an unrunnable ffmpeg command in the exact #378 scenario. #408 fixes the decode/filter-graph mismatch and covers the progressive path too.

PR #405, opened 2026-09-01
feat(collections): image configurator

lostb1t:


jonas-thuvesson-winberg:
## 🐛 Bugs? 

1. Update preview doesn't change the posters even if you have changed the media filter to include other items. If you hit save and then re-open the collection (edit), you get the updated posters.

2. The font picker is empty (no selection) when you hit save and then re-open (edit).

3. If you update preview with "Streaming Logo" selected, but no provider, a text "Streaming" appears on the card.

## ✨ Suggestions

1. Would be a nice feature to be able to auto-fit the title (text) inside the card (meaning the font size adjusts). It should probably be optional though, as having control of font-size is also nice if you want to have a unified look (same size for all your collections). Current implementation is fine though!

2. I understand if you don't want to do this one: With the added options the modal/dialog window becomes loooong. Would look better if it was using more horizontal space instead (preferrably with elements being able to rearange themselves using grid/flex so that it _can_ be narrow on mobile).

---

Will write if I find something more!

jonas-thuvesson-winberg:
Buttons (and the slider) sometimes overflow instead of wrapping.

image

PR #411, opened 2026-09-02
fix(playback): use hev1 when HEVC parameter sets are in-band

A-wes:
Hello! So while my 10-bit streams weren't working I tried to use 8-bit alternatives, only to discover that they seem to crash clients too, but this time due to other issue. Unfortunately I do not have deep codecs knowledge, unlike claude, so submitting it as is. Both code in this PR and explanation below are written by Claude. I did verify this fix worked for WEB-DL playback on my NVIDIA Shield 

When stream-copying HEVC into fMP4/HLS, `-tag:v hvc1` asserts that the sample entry carries a complete `HEVCDecoderConfigurationRecord` with VPS, SPS and PPS out-of-band. Some sources — typically WEB-DL repackages from third-party muxers — report extradata that is a bare header-only record with `numOfArrays = 0`, leaving parameter sets in-band in the bitstream. Forcing `hvc1` on those writes an empty 8-byte `hvcC` box, which ExoPlayer-based clients cannot initialise a decoder from. `hev1` degrades gracefully in the same situation.

`ffprobe` reports `extradata_size` in its default `-show_streams` output, and per ISO/IEC 14496-15 §8.3.3.1.2 the fixed header of a `HEVCDecoderConfigurationRecord` is exactly 23 bytes, so `extradata_size > 23` is a sound, parsing-free signal for "has real out-of-band parameter sets" — no need to decode the extradata itself.

`probe.rs` captures `extradata_size` per HEVC video stream and derives `hevc_params_out_of_band`, exposed as an additive `remux`-namespace field on `MediaStream` (`MediaStreamRemuxExt`, following the existing extension pattern). It is threaded through `TranscodeSession` / `TranscodeParams` / `ProgressiveTranscodeParams` to both `-tag:v` sites via a new `hevc_copy_tag()` helper.

The helper defaults to today's `hvc1` whenever the signal is unknown or positive, so a source that works now cannot regress.

Note: `-bsf:v extract_extradata` is a commonly cited fix for this symptom and was deliberately not used — it requires Annex-B NAL framing and fails with "No start code is found" against Matroska's length-prefixed HEVC packets, which would break every HEVC copy from an MKV source.

### DIFF
--- a/crates/remux-sdks/src/remux/mod.rs
+++ b/crates/remux-sdks/src/remux/mod.rs
@@ -3057,6 +3057,13 @@ pub struct UpdateUserPassword {
     pub reset_password: Option<bool>,
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
+pub struct MediaStreamRemuxExt {
+    /// HEVC only: whether extradata carries VPS/SPS/PPS out-of-band. Picks
+    /// between the `hvc1` and `hev1` sample-entry tags on stream copy.
+    pub hevc_params_out_of_band: Option<bool>,
+}
+
 #[dto]
 pub struct MediaStream {
     pub aspect_ratio: Option<String>,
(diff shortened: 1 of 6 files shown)


lostb1t:
Superseded by #413, which takes the signal from the client's DeviceProfile instead of the source's extradata. I kept your CODECS fix.

Would appreciate you testing #413 on your Shield if you get a chance

PR #416, opened 2026-09-02
fix(items): return MediaSources when an item is fetched by MediaSourc…

A-wes:
Hi, while testing HEVC playback fix on my shield ran into an issue when picking alternative media source results in client crash. Below is claude's explanation of what's happening after reading server logs and org.jellyfin.androidtv client logs uploaded after crash. I verified crash does not happen after adding this

Android TV refetches an item by MediaSource id when the user picks a different version, and `item_for_user` already answers that: it resolves the Stream row and sets `sources` to the row itself.

`db_media_to_item` then discarded that work. Its `MediaSources` block is gated on `Movie | Episode | Track`, so for a `Stream` row the field was never populated and `skip_serializing_none` omitted it from the response entirely. `GET /Items/{mediaSourceId}` returned 200 with `Type: "Video"` and no `MediaSources` at all.

A client that reads `MediaSources` off that response with no `PlaybackInfo` round-trip therefore gets null. Jellyfin for Android TV dereferences it unguarded in `PlaybackController.buildExoPlayerOptions` and crashes:

    java.lang.NullPointerException: Attempt to invoke virtual method
    'java.util.List org.jellyfin.sdk.model.api.MediaSourceInfo.getMediaStreams()'
    on a null object reference

The app relaunches, retries the same item and crashes again, so playback never recovers.

Add `Stream` to the gate so the sources `item_for_user` resolved are actually serialised.

Fuller log just in case:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.List org.jellyfin.sdk.model.api.MediaSourceInfo.getMediaStreams()' on a null object reference
	at org.jellyfin.androidtv.ui.playback.PlaybackController.buildExoPlayerOptions(PlaybackController.java:521)
	at org.jellyfin.androidtv.ui.playback.PlaybackController.play(PlaybackController.java:486)
	at org.jellyfin.androidtv.ui.playback.PlaybackController.play(PlaybackController.java:392)
	at org.jellyfin.androidtv.ui.playback.CustomPlaybackOverlayFragment.onActivityCreated(CustomPlaybackOverlayFragment.java:300)
	... (Fragment/Handler/Looper/ActivityThread frames, standard Android crash unwind)
```

### DIFF
--- a/crates/remux-server/src/api/models.rs
+++ b/crates/remux-server/src/api/models.rs
@@ -1034,10 +1034,13 @@ pub fn db_media_to_item(media: db::Media, hide_sources: bool) -> BaseItemDto {
     item.external_urls = external_urls;
 
     // several dlients require at least one stream
+    // `Stream` is here because Android TV refetches an item by MediaSource id
+    // when switching versions; without it that response has no MediaSources.
     if !hide_sources
         && (media.kind == db::MediaKind::Movie
             || media.kind == db::MediaKind::Episode
-            || media.kind == db::MediaKind::Track)
+            || media.kind == db::MediaKind::Track
+            || media.kind == db::MediaKind::Stream)
     {
         item.can_download = Some(true);
         item.media_sources = match media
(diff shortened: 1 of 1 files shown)


lostb1t:
i cannot replicate this. bug reports are appreciated but please just open an issue instead of posting AI slop prs

what jellyfin app version are you on? and what remux version

A-wes:
Hi, sorry, I will open issues in the future, feel free to close the PRs (I was creating them as a reference to what seems to be fixing the issue, not expecting them to merged. I understand that it is annoying).
App version is Jellyfin for Android TV 0.19.10, remux was built from `fix/hevc-tag-from-device-profile` branch. But I believe I had encountered this issue on earlier versions too

lostb1t:
ar you playing from the details
page? are you using groups?

A-wes:
I do not have groups set up in remux. Also it looks like this is happening when I select the full season packs (big files 10+GBs) They are often the only available source AIOStreams returns. Can groups solve this?

lostb1t:
season packs? are you using p2p or debrid? and are you using aiostreams

A-wes:
Yes I am using aiostreams + realdbrid. Some of the results AIOStreams returns are single episodes some are entire season that incudes the episode. Issue seems to be occurring in the latter. My bad for not noticing it earlier, in some cases season packs are the only results returned, so I did not notice a difference. Screenshot below is how it looks like on RD side 
{A13369BC-20F2-4905-9F4D-14BFE94DFCCA}

lostb1t:
k ill have to test that. didn't know aiostreams supported packs

Issue #436, opened 2026-09-06
Add settings for changing stream cache TTL and clearing cache


jonas-thuvesson-winberg:
Would be nice to be able to control the stream caching some more. The ability to change how long streams are cached, and clearing all cached streams.

PR #439, opened 2026-09-06
feat(dashboard): add stream cache setting and task

jonas-thuvesson-winberg:
See (#436).

* Add stream cache ttl setting (default 60 seconds).
* Add clear all cached streams task.

Disclosure: Codex was used (under supervision).


lostb1t:
i don't want it ti make it another setting. its fina as is

task is bit of an overkill. cant remember if i added an autoclean to the refresh library (like streams that are a week old) if not lets put it there

jonas-thuvesson-winberg:
MOAR SETTINGS! Ok, fair enough.

jonas-thuvesson-winberg:
How about an env var? 😬

lostb1t:
why do you want to change it? what issue you are running into

jonas-thuvesson-winberg:
Sometimes you want to test a formatter change or something, and the jf web > "refresh metadata > scan for new and updated files" rarely works. Would be nice to just force wipe it with the press of a button. But it's not a big thing. I just like to have things configurable (and this was a setting in gelato).

PR #464, opened 2026-09-10
fix(playback): honor the probe fallback on the stream request that follows

brokenthumbs:
PlaybackInfo probes the first source. If that times out it falls back to the next one and returns it, with `MediaSources[0].Id` set to the item id as usual. A client that auto-plays then requests `/videos/{id}/stream` with `MediaSourceId` equal to the item id. `dispatch_lookup` reads that as "first source" and serves the one that just failed. The request hangs until the upstream timeout and returns 500. Choosing the second source by hand works, because then the client names it directly.

Stream groups have the same problem: a group request answers with the group id, and the stream request resolves the group to its first candidate regardless of what the probe found.

Fix: after probing, `save_probe_fallback` records the stream that actually probed, keyed by the play session id and the id the client will echo back (the item id, or the group id). The stream handler looks that record up with the same key and uses it. A specific stream named by the client has no record and still wins. Keyed on the play session rather than the device because every client echoes it on the stream URL; not all send a device id there. Unit tests cover auto-play, group requests and specific streams. Full suite green.

`GET /videos/{id}/stream?MediaSourceId={id}` after PlaybackInfo had fallen back, first source unreachable:

```
before: 500 after 30 s   after: 206 in 0.4 s
```

Same check with a stream group defined and the group requested by its id, first candidate timing out: the stream request served the fallback stream (verified by file size), 206 in 0.3 s.

Written with Claude Code; I reviewed and tested it on my own instance.

### DIFF
--- a/crates/remux-server/src/api/playback.rs
+++ b/crates/remux-server/src/api/playback.rs
@@ -846,20 +846,6 @@ fn ext_from_descriptor(descriptor: &crate::stream::StreamDescriptor) -> String {
     }
 }
 
-/// Which id the stream request resolves: a specific stream when the client
-/// named one, otherwise the probe fallback remembered for the play session
-/// (if any), otherwise whatever the client sent (item id or nothing).
-fn stream_lookup_id(
-    item_id: Uuid,
-    media_source_id: Option<Uuid>,
-    probe_fallback: Option<Uuid>,
-) -> Option<Uuid> {
-    match media_source_id {
-        Some(sid) if sid != item_id => Some(sid),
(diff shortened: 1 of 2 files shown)


lostb1t:
did you also test this with stream groups enabled?

brokenthumbs:
> did you also test this with stream groups enabled?

good call out! updated code to also handle fallback for groups. fallback is now keyed by play session and id (group or item). verified on local instance with group defined: first candidate timing out, the stream request served the fallback stream.

PR #465, opened 2026-09-10
fix(search): return the stored row's id for remote results that exist locally

brokenthumbs:
Remote search results get a fresh id on every request, held in the in-memory store for an hour. Searching the same title repeatedly returns a different series id each time, none of them the stored series row, even though the episodes listed under each one name that row as their series. A client that keeps the id it was given, for next-episode autoplay, a continue-watching entry or the show page, comes back after the hour or after a restart and gets 404 on `/shows/{id}/seasons`: autoplay stops at the episode end, the entry opens blank, the show has no source until the client searches again.

Fix, two layers. `Media::try_from(Meta)` now derives the id from the result's external ids the way a stored row's id is derived, so the first search for a title and every search after it agree before anything is stored. For rows that ended up stored under a different id (matched through another provider's id), `Media::adopt_existing_ids` rewrites the result to the stored id, one query per kind with the usual id priority applied in memory. A result with no identity and no match keeps a random id. Two unit tests. Full suite green.

Series search for a title that already exists locally, four requests:

```
before: four different ids, none the stored row
after:  the stored row's id every time
```

Written with Claude Code; I reviewed and tested it on my own instance.

### DIFF
--- a/crates/remux-server/src/db/media.rs
+++ b/crates/remux-server/src/db/media.rs
@@ -1978,6 +1978,7 @@ pub enum MediaError {
 // doesn't coerce across storage classes in a bare `=`), so integer-valued
 // fields must bind as an integer, not a stringified one. Used by
 // `Media::find_by_external_ids` and `Media::resolve_ambiguous_external_id`.
+#[derive(Clone, PartialEq)]
 enum IdValue {
     Text(String),
     Int(i64),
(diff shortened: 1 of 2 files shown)


crates/remux-server/src/db/media.rs
```diff
@@ -2773,6 +2773,30 @@ impl Media {
...
+    /// Point remote results (search, catalog) that already exist locally at
+    /// their stored row: same kind, matching external id.
+    ///
+    /// Remote results are minted with a fresh id per request and live in the
+    /// in-memory store for an hour. A client that keeps such an id, for
+    /// next-episode autoplay or a continue-watching entry, is left with a
+    /// dead one after that, or after a restart, and gets 404 on
+    /// `/shows/{id}/seasons` while the stored series is fine. Only results
+    /// with a matching row are rewritten; unknown items keep their id.
+    pub async fn adopt_existing_ids(db: &SqlitePool, items: &mut [Media]) {
```
lostb1t, line 2786: an query per item is pretty inefficient. Bulk load items.

lostb1t (changes requested):

crates/remux-server/src/db/media.rs
brokenthumbs: good catch, thanks. updated: now using one query per kind, priority applied in memory.

lostb1t:
but this doesnt solve the first insert no?

the uuid will be random
of the first search because there isnt any local version.

brokenthumbs:
> but this doesnt solve the first insert no?
> 
> the uuid will be random of the first search because there isnt any local version.

ah, sorry, i only tested the initial fix with titles already stored. updated code now derives the id from the external ids, the same way stored rows get theirs, so the first search and every one after agree before anything exists.

PR #467, opened 2026-09-11
fix: stop opendal-local .strm scan from deadlocking on URL collisions

northernpowerhouse:
For .strm entries, the row `id` is derived from the file's own filesystem path while the stored `path` column holds the URL read from inside the file. Two different .strm files can therefore produce different ids but the same path (e.g. after VOD2MLIB rewrites a title's proxy URL), which a plain INSERT can't reconcile via `ON CONFLICT(id)` and instead trips the UNIQUE(addon_id, path) index.

That error was propagated with `?`, aborting the whole scan before reaching prune_stale_paths — so the colliding stale row was never cleaned up and every subsequent scan hit the identical collision at the identical point, permanently wedging the addon at 0 indexed files.

Also fixes fetch_existing_imdb being called with the filesystem path instead of the stored URL for .strm entries, which meant the imdb_id-reuse lookup could never match anything.

### DIFF
--- a/crates/remux-server/src/addons/opendal.rs
+++ b/crates/remux-server/src/addons/opendal.rs
@@ -1628,19 +1628,22 @@ async fn scan_addon(
             // has since been renamed/regenerated (e.g. a `.strm` whose URL rotated).
             // `id` and `path` come from different sources for `.strm` entries (fs path
             // vs. the URL read from the file), so `ON CONFLICT(id)` can't reconcile it.
-            // Skip this one row instead of aborting the whole scan via `?` — bailing
-            // out here would also skip `prune_stale_paths` below, so the stale row
-            // would never get cleaned up and every future scan would hit the same
-            // clash again.
+            // Skip only that specific violation instead of aborting the whole scan via
+            // `?` — bailing out here would also skip `prune_stale_paths` below, so the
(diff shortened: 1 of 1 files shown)


northernpowerhouse:
Addressed both review comments in 309241a:

- **Error handling**: narrowed the insert-error match to only swallow `sqlx::Error::Database` where `is_unique_violation()` is true (i.e. the specific `UNIQUE(addon_id, path)` clash this fix targets). Every other error — connection loss, disk I/O, FK violation, etc. — now propagates via `return Err(e.into())` instead of being logged and skipped, so the scan can no longer report success while `prune_stale_paths` acts on an incomplete `seen_ids`.
- **Test coverage**: added `opendal_strm_rename_recovers_from_stale_row_collision`, which reproduces the actual bug scenario from the PR description — index a file, delete it and add a differently-named file pointing at the same proxy URL (simulating VOD2MLIB regenerating a `.strm`), then rescan. It asserts the first rescan doesn't abort and still prunes the now-stale row despite the insert collision, and that a second rescan then cleanly indexes the current file with its IMDb id intact. The original two-simultaneous-files test is kept as-is since it still usefully covers the "never abort" guarantee on its own.

Also rebased onto the latest `main` merge (the 0.31.0 release + the `media_relations` dedupe fix) since the branch had diverged.

PR #483, opened 2026-09-13
∞

moodiness:
∞


lostb1t:
web patching is meant for minimum changes to the web ui. If you want more changes like this i suggest using JS injection from the branding page

PR #486, opened 2026-09-14
∞

moodiness:
∞


moodiness:
Pushed `36ed4366`, which reworks link management to be admin-issued and adds the UI.

**Admin issues, admin distributes.** Link endpoints now require `AdminSession` and are keyed by user id:

| Method | Route |
|---|---|
| `GET` | `/remux/calendar/links` |
| `POST` | `/remux/calendar/links/{user_id}` |
| `POST` | `/remux/calendar/links/{user_id}/rotate` |
| `DELETE` | `/remux/calendar/links/{user_id}` |

Issuing is **idempotent**: re-reading someone's URL to send it on returns the existing link. Rotating it instead would silently break the subscription they already added to their calendar app, which is not something an admin fetching a URL should be able to do by accident. Rotation is a separate explicit endpoint.

**Storage changed from digest to plaintext.** The admin UI has to display the token, so it must be readable. This follows `api_keys`, which stores `access_token` the same way for the same reason. Consequence, stated plainly: `user_calendar_links` is credential material — a leak yields working feed URLs. Each grants read access to one user's calendar and nothing else, and is individually revocable.

**UI:** a Calendar page under `Content`, listing each link with owner, absolute URL, copy, rotate and revoke. Generating only offers users without a link.

`cargo test -p remux-server`: **699 passed, 0 failed**. `cargo fmt --check` clean on `remux-server`, `remux-sdks` and `remux-dashboard`; `remux-dashboard` also checked against `wasm32-unknown-unknown`.

Two new tests worth noting:

- `link_endpoints_are_admin_only` — a non-admin session is refused on all four routes, including its own link
- `issuing_is_idempotent_and_rotating_replaces_the_url` — re-issuing returns the same URL; rotating invalidates the previous one

Non-admin users still have no way to fetch their own link; they depend on an admin sending it. That is the intended model here, not an oversight.

moodiness:
Pushed `4b236b8a`. I measured the query on a real library instead of leaving the note speculative, and the earlier "unmeasured" caveat turned out to understate it.

**Setup:** 212k media rows accumulated by a live sync, 30 favorited series, 761 events in the window.

| | before | after |
|---|---|---|
| window scan | 74 ms | **2 ms** |
| full feed request | ~460 ms | **~170 ms** |

The cause is that neither `idx_media_released_at` nor `idx_media_digital_released_at` can serve `COALESCE(released_at, digital_released_at)`, so SQLite fell back to `SEARCH media USING INDEX idx_media_kind_avail_sentinel (kind=?)` and filtered every episode row. Indexing the expression switches it to `SEARCH ... (kind=? AND <expr>>? AND <expr><?)`. Output is byte-identical before and after verified by diffing the serialised feeds.

Two other things the measurement surfaced:

- **N+1 on series titles.** `Media::get_by_id` issues two statements per call the row, then its images, which the calendar never reads. A followed soap opera with 80 in-window episodes cost 160 round trips to read one title 80 times. Now one batched `get_by_ids` over distinct series ids.
- **`MAX_EVENTS` was a per-kind cap, not a global one.** Movies and episodes were queried separately although both are bounded by the same premiere expression, so the 5000 limit could yield up to 10000 events. They now share one query.

Also moved the window's end bound to the last representable nanosecond of the day: an item stored at `23:59:59.5` on the final date was being dropped.

`cargo test -p remux-server`: **699 passed, 0 failed**. `cargo fmt --check` clean.

Note on the earlier PR text: it said materialisation via a task would be the next step if this proved slow. Having measured it, an index alone is sufficient at this scale and the task idea is not warranted I'd drop that suggestion rather than leave it as a standing recommendation.

PR #523, opened 2026-09-20
Fix/refframes apply conditions

### LINKED ISSUE #521: Bug: Direct play is blocked for H.264 files because remux ignores conditional RefFrames rules and never reads refs from ffprobe
## Summary

Some clients send conditional `RefFrames` rules for H.264. Remux ignores the conditions, and it also never reads the `refs` value from ffprobe. Files with no stored `RefFrames` fail the rule, get labelled `VideoCodecNotSupported`, and lose direct play. The device decodes them fine.

## What the client sends

Moonfin for Android TV 2.5.1 puts these in `CodecProfiles` for h264:

- `RefFrames <= 12`, only apply if `Width >= 1200`
- `RefFrames <= 4`, only apply if `Width >= 1900`

The "only apply if" part is `ApplyConditions`. It's how Jellyfin-style profiles scope a rule so small files aren't held to the strict limit.

## What remux does wrong

1. `CodecProfile` has no `ApplyConditions` field, so serde drops it. Every rule applies to every H.264 file. A 720p file with 5 ref frames fails the `<= 4` rule, which was meant for 1900+ wide files.
2. `FfprobeStream` in `playback/probe.rs` has no `refs` field. ffprobe reports `refs`, but it's discarded, so `ref_frames` stays empty for anything probed this way. The rules are sent with `IsRequired: true`, so an empty value counts as a failure.
3. The failure is reported as `VideoCodecNotSupported`, because `CodecProfile::check_reasons` sends any unrecognised property (`RefFrames` included) to that catch-all. The codec is fine, and the wrong reason makes this very hard to debug.

Both 1 and 2 are still present on current `main` (`5c907e8c`).

## Effect

Any H.264 file without a stored `RefFrames` loses direct play. It gets an HLS remux, or a full re-encode if video transcoding is enabled. In my DB, 210 of 517 H.264 probes (about 40%) have no `RefFrames`.

I think this became visible after #406 ("evaluate codec profiles for every direct play candidate"), since codec profile rules now gate every direct play decision instead of only a fallback path. I haven't tested a build from before it.

## Repro

- Client: Moonfin for Android TV 2.5.1 on an Ugoos SK1. The client reports AVC level 5.2 and a hardware H.264 decoder, and it does decode the stream.
- Server: `ghcr.io/lostb1t/remux:latest`, image built 2026-09-19.
- File: H.264 High, level 4.0, 1920x800, 8-bit, progressive, mkv. `ffprobe` reports `refs: 1`.
- Remux's stored probe for the file has no `RefFrames`.
- Client log: `transcode (VideoCodecNotSupported)`, with `offeredByServer: false`.

I ran the client's real profile against the stored probe data in `check_direct_play`:

| Case | Result |
|---|---|
| `RefFrames` absent | `VideoCodecNotSupported(property=RefFrames condition=LessThanEqual value=12 actual=(unknown))` |
| `RefFrames = 4` | direct play |
| `RefFrames = 5` | fails on the `<= 4` rule, even for files that rule shouldn't cover |
| `RefFrames` rules removed | direct play |

The deployed binary has no `ApplyConditions` string in it, so the field really is missing at runtime.

## Where `RefFrames` values come from

The CHANGELOG says `ref_frames` was added to the remuxdb mediainfo payload. That probably explains why some streams have it and others don't: remuxdb supplies it, remux's own ffprobe path doesn't. I haven't traced this in the code, so treat it as a guess.

## Proposed fix

1. Read `refs` in the ffprobe parser and set `ref_frames` on video streams. About two lines. This alone fixes the file above.
2. Add `apply_conditions: Vec<ProfileCondition>` to `CodecProfile` (`#[serde(rename = "ApplyConditions")]`). In `check_codec_profiles`, skip a profile unless all of its apply conditions hold for the stream.
3. Report accurate reasons instead of the catch-all. Jellyfin has `RefFramesNotSupported`, `VideoLevelNotSupported` and a resolution reason, so use those (or add them). `VideoCodecNotSupported` should mean the codec. #291 already notes this catch-all can force unnecessary video encoding for non-codec failures.
4. Re-probe streams already stored without `refs`, so existing libraries are fixed without waiting for a refresh.

## Related

- #406: made codec profile rules apply to every direct play candidate.
- #291: notes the `VideoCodecNotSupported` catch-all forcing unnecessary H.264 encoding.

northernpowerhouse:
Fixes #521.

Some clients (in my case Moonfin on Android TV) send RefFrames limits that are scoped with ApplyConditions (`RefFrames <= 4` only if `Width >= 1900`, and so on). Remux dropped ApplyConditions, so every H.264 file was held to every limit. It also never read `refs` from ffprobe, so files with no stored RefFrames failed the IsRequired rule and lost direct play. The failure was reported as VideoCodecNotSupported, which made it hard to track down.

What this changes:

- `CodecProfile` gets `apply_conditions`; a profile is skipped unless all of   its apply conditions hold for the stream. `hevc_copy_tag` respects it too.
- `refs` is read from ffprobe into `ref_frames`.
- RefFrames, VideoLevel and Width/Height failures now report  `RefFramesNotSupported`, `VideoLevelNotSupported` and `VideoResolutionNotSupported` instead of the VideoCodecNotSupported catch-all. They still trigger a video transcode like before, so behaviour doesn't change apart from the reason shown.
- A cached ffprobe result for an H.264 stream with no `ref_frames` is treated as stale and re-probed on the next playback request.

The commit was drafted by Claude. I revised it and tested it.

## Testing

- Added unit tests in `device_profile.rs` and `probe.rs`, including the exact case from the issue (Moonfin's RefFrames rules against a 1920-wide stream with and without RefFrames).
- `cargo test -p remux-server -p remux-sdks` passes. 
- Built an image from this branch and ran it against my real library. Sent PlaybackInfo with the real SK1 Moonfin profile: `refs` now gets stored on fresh probes, scoped rules are skipped when the scope doesn't match, and the new reasons show up correctly.

## Not covered

- The re-probe only covers streams whose stored probe came from ffprobe. Entries that came from remuxdb without `ref_frames` still fail an IsRequired RefFrames rule until remuxdb has the value.


lostb1t:
superseded by https://github.com/lostb1t/remux/pull/524

